PR #7, opened 2020-02-11
Add Lambda function redirecting URIs to S3 file objects

nathanegillett:
Includes commits relating to redirecting URI origin requests to S3 file object keys.


cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,37 @@
+from datetime import datetime
+
+from chexus import Client, TableItem
```
rohanpm: I didn't know that you intend to integrate chexus here, I thought that was targeting only the publish side.

If you want to use chexus here, can you please donate it to release-engineering organization first, nominate CODEOWNERS, start using code review etc?  I would have asked for that around Phase II originally, but if you want to use it in the lambdas it should happen earlier.

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,37 @@
...
+def lambda_handler(event, context):
+    _ = context
+
+    # Get uri from origin request event
+    request = event["Records"][0]["cf"]["request"]
+
+    client = Client()
+
+    # Get DB table items with matching web_uri key
+    search_results = client.search(
```
rohanpm: This doesn't quite seem right because we're searching for *all* matches on the URI, and then filtering client-side for the right one to use.  Can't the DB do such filtering?  Over time, there will be hundreds to thousands of items for the same URI (thinking of files updated with high frequency like repomd.xml).

As a side note, the DB limits for fetching are not very high (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-api). I'm not sure what operation we are doing here but it would either be BatchGetItem (100 items) or Query (1MB) I guess, both of which it seems we could hit if doing an insufficiently filtered query.  I haven't looked into whether you already added something to chexus for pagination.

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,37 @@
...
+        item=TableItem(web_uri=request["uri"]),
+        table_name=TABLE_NAME,
+        region=TABLE_REGION
+    )
+
+    items = [TableItem(**i) for i in search_results["Items"]
+             if i.attrs["from_date"] <= datetime.utcnow().date()]
```
rohanpm: Surprised we are only doing date here.

In scope of this issue we expect to work on DB schema a little, so this seems a reasonable place to point it out: I'm pretty sure these should be full ISO8601 timestamps down to the second, and not just dates.  Using only date makes it impossible to distinguish two publishes of the same piece of content within the same day, which is common.

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if items:
+        # Sort the items by date
+        items.sort(key=lambda i: i.from_date)
+
+        # Redirect the request uri
+        request["uri"] = items[-1].attrs["object_key"]
+    else:
+        # Invalid request
+        pass
```
rohanpm: I think we need to explicitly give 404 here for reasons:

- if we don't, then users who know about our S3 layout (which is anyone who cares to look, given the code is public) will be able to request objects by checksum.  It's not an intentional feature, so we should block it.
- if the request falls to S3 and generates a 404 there, I think that'll cost us slightly more money, as we still need to pay for the S3 request.

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,37 @@
...
+
+    items = [TableItem(**i) for i in search_results["Items"]
+             if i.attrs["from_date"] <= datetime.utcnow().date()]
+
+    if items:
+        # Sort the items by date
+        items.sort(key=lambda i: i.from_date)
+
+        # Redirect the request uri
+        request["uri"] = items[-1].attrs["object_key"]
```
rohanpm: I have a question about logging, which might need some looking into.

If something goes wrong here - e.g. imagine that the item from the DB is missing object_key somehow, so this raises KeyError - do we get meaningful logs somewhere with the full backtrace?  Or do we need to surround the entire thing with an exception handler and a call to a logger?

It's fine to leave this question for later though.

nathanegillett:
So, yeah, this PR is just a draft, @rohanpm. I wanted to push what I had at EOD yesterday and add you in case you had any notes.

cdn_lambda/functions/map_to_s3.py
nathanegillett: Yes, I agree. This is just the way it would have to be implemented if using chexus in it's current state, as it only queries exact matches.

I'll build a boto3 query here to retrieve only the latest item.

cdn_lambda/functions/map_to_s3.py
nathanegillett: Ah, I hadn't considered multiple releases in a single day. Thanks.

cdn_lambda/functions/map_to_s3.py
nathanegillett: Ah, yes, this was a placeholder for the else statement. Yeah, this'll return a 404 response.

cdn_lambda/functions/map_to_s3.py
nathanegillett: No, I do think we'll need to catch things like that here. Our tooling can do everything to ensure it abides by the schema but I think there's still room for invalid content to find it's way onto the DB, and this is that content's first stop.

cdn_lambda/functions/map_to_s3.py
nathanegillett: I think we don't need to use chexus here, but I've considered the possibility of transferring it to release-engineering.

If the code does go to release-engineering, I would prefer creating a new project there, initializing it with the current chexus code and prehaps changing the name, as I was never all that fond of the checksum-nexus portmanteau.

nathanegillett:
TO DO: "The DynamoDB and S3 schemas are covered in the repo's published docs"

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+from datetime import datetime
+
+import boto3
+
+LOG = logging.getLogger("map-to-s3-lambda")
+
+
+def lambda_handler(event, context):
+    _ = context
```
rohanpm: What does this do?

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,61 @@
...
+    LOG.info("Querying '%s' table for '%s'...", table_name, web_uri)
+
+    query_result = boto3.client("dynamodb", region_name=table_region).query(
+        TableName=table_name,
+        Limit=1,
+        ScanIndexForward=False,
+        KeyConditionExpression="web_uri = :u and from_date <= :d",
+        ExpressionAttributeValues={
+            ":u": {"S": web_uri},
+            ":d": {"S": str(datetime.utcnow())},
```
rohanpm: The default str formatting of datetime here doesn't seem ideal, it doesn't seem to be exactly one of the standard ISO8601 formats (e.g. the same format Pulp uses).  I played around with it a bit and I think the best way to get our timestamps is like this:

```
datetime.now(timezone.utc).isoformat(timespec='milliseconds')
```

- gives a value like 2020-02-13T04:20:40.444+00:00 which follows ISO8601 standard, and includes timezone so there's no ambiguity
- datetime.now(timezone.utc) is recommended over datetime.utcnow (according to https://docs.python.org/3.8/library/datetime.html)
- The default behavior of str/isoformat is "timespec='auto': Same as 'seconds' if microsecond is 0, same as 'microseconds' otherwise."  This seems to be undesirable, as datetimes which happen to land on a microsecond=0 (very small chance) will happen to have a different length from all other datetimes.  So there will be a small number of timestamps which use a different format for no real reason.

cdn_lambda/functions/map_to_s3.py
```diff
@@ -0,0 +1,61 @@
...
+    if query_result["Items"]:
+        LOG.info("Item found for '%s'", web_uri)
+
+        # Update request uri to point to S3 object key
+        latest_item = query_result["Items"][0]
+
+        try:
+            request["uri"] = latest_item["object_key"]["S"]
+            return request
+        except KeyError:
```
rohanpm: A few points here:

1. The possible case I mentioned earlier where object_key could be missing was just an example - in reality, there's unlimited possible issues, right?  For instance, the query to DynamoDB could fail with a connection error.  If we do have to implement our own logging for errors, it seems like it needs to surround everything which could fail.  I was thinking more like a `@decorator` which can easily catch everything, rather than adding try/except blocks.

2. I don't think it's a good idea to try to make our own error message in this way - I mean, just because you got a KeyError, you don't really *know* that 'object_key' was the missing key.  And the KeyError exception itself formats to a string telling you which key was missing.  So, adding this kind of message duplicates info which is already available but makes it possibly inaccurate (as you might have guessed wrong why a KeyError was raised).

3. Should we explicitly return 500 actually, or just let the exception propagate?  I haven't investigated what happens if an exception is allowed to propagate from a CloudFront lambda, but it might already be giving us some nice features for free - like generating a unique ID for the failure which can be cross-referenced in other logs.  This is something missing from your return 500.  Also, you're allowing the client to see the entire content of the DynamoDB record, we probably shouldn't allow that in general, we might be stashing metadata in there that we don't want customers to see.

So putting that all together:

- If we are responsible for logging our own exceptions, then the best way is probably to wrap the function with a decorator, which calls LOG.exception("Internal error during request").  LOG.exception logs the full backtrace.
- I suggest that decorator can also re-raise the exception after logging, rather than explicitly returning 500, and then we should have a look what happens in CloudFront in that case.
- If you think it's important for the latest_item to be logged in case of error, that can be included via try/except *as well as* the decorator, for instance:

```
latest_item = query_result["Items"][0]
try:
  ... # code using latest_item
except:
  LOG.exception("Exception occurred while processing %s", latest_item)
  raise
```

tests/functions/test_map_to_s3.py
```diff
@@ -0,0 +1,134 @@
...
+import os
+import mock
+import json
+
+
+@mock.patch("boto3.client")
+@mock.patch("cdn_lambda.functions.map_to_s3.datetime")
+def test_map_to_s3(mocked_datetime, mocked_boto3_client):
+    mocked_datetime.utcnow.return_value = "2020-02-12 20:55:06.372878"
+    mocked_boto3_client().query.return_value = {
```
rohanpm: Have you seen moto library? https://github.com/spulec/moto

I think we should give that a try for the tests.  I'm really not a fan of the mocking approach here, because tests using these mocks are very weak.  mocked_boto3_client().query.assert_called_with(absolute_garbage) is going to pass if both the test and the code under test use the same garbage.

As I understand it, with moto we could use the regular boto3 API to insert records into a DynamoDB, then the queries in our lambda code should work normally instead of going into a mock which doesn't do anything.  We can then get by just with testing the outcome of the lambda.

That being said, I don't know how complete moto's DynamoDB backend is, so someone would have to give it a try and see if it's workable.

tests/functions/test_map_to_s3.py
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    env_vars = {"DB_TABLE_NAME": "test_table", "DB_TABLE_REGION": "us-east-1"}
+    event = {
+        "Records": [
+            {"cf": {"request": {"uri": "www.example.com/content/file.ext"}}}
+        ]
+    }
+
+    with mock.patch.dict(os.environ, env_vars):
```
rohanpm: There's actually a built-in fixture to help with this - https://docs.pytest.org/en/latest/monkeypatch.html

```
monkeypatch.setenv("DB_TABLE_NAME", "test_table")
monkeypatch.setenv("DB_TABLE_REGION", "us-east-1")
```

They'll be unwound automatically at the end of the test.  I leave you to decide if this is better or worse than the current approach.

cdn_lambda/functions/map_to_s3.py
nathanegillett: The context variable is unused otherwise but I believe that the handler must be able to accept it.
So I guess all this is really doing is `pylint: disable=unused-argument`.

tests/functions/test_map_to_s3.py
nathanegillett: I think I prefer the current approach as it does the same thing fewer imports and less code.

tests/functions/test_map_to_s3.py
nathanegillett: While it looks promising, moto doesn't "just work" apparently. I can't even successfully put an item into a DB table. Also, looking through the issues on Github, it seems there are a lot of behavioral inconsistencies.

cdn_lambda/functions/map_to_s3.py
nathanegillett: So, I've broadened the except and logged a more generic exception like you've mentioned, raising the caught exception. I'll experiment on this a bit and see if we need to do any other handling.

PR #11, opened 2020-02-20
Fix map_to_s3 configuration

### LINKED ISSUE #8: map_to_s3 function depends on non-existent enviroment variables
#### Summary
Lambda@Edge functions does not support environment variables, yet map_to_s3 depends on DB_TABLE_NAME and DB_TABLE_REGION. Another method of injecting these variables must be implemented.

#### Issue Type
- Bug Report

#### Steps to reproduce
Attempt to deploy map_to_s3 function with any environment variables.

#### Actual results
Function creation is rejected by AWS Lambda

#### Expected results
Function creation is successful and the function can access determine the correct table to query for mapping.

#### OS / Environment
Lambda@Edge, Python3.7

### LINKED ISSUE #10: '/' missing from the uri returned to cloudfront
#### Summary
The uri in the request passed to CloudfFont should begin with a '/'
https://github.com/release-engineering/cdn-lambda/blob/master/cdn_lambda/functions/map_to_s3.py#L44

#### Issue Type
- Bug Report

#### Steps to reproduce
Attempt to access a valid path that exists in the ddb + object mappings table.

#### Actual results

> ERROR Validation error: Lambda function result validation failed, the specified URI is in an invalid format, uri must start with '/'.

User receives a "502" error from CloudFront after attempting to access cloudfront-domain.cloudfront.net _objectkey_): 

> The Lambda function result failed validation: The specified URI is in an invalid format.

#### Expected results
The uri in the request that is passed to CloudFront begins with '/'. The user is redirected to object key (i.e., attempts to access cloudfront-domain.cloudfront.net/_objectkey_))

#### OS / Environment
Lambda@Edge, Python3.7

#### Additional Information

nathanegillett:
Environment variables are not supported by Lambda@Edge, so replace their
usage with a configuration file. A JSON format file was chosen to
prevent additional dependencies.

Also, the transformed URI is now prepended with a forward-slash, "/", to
correct the path to the file object.

Tests were updated to reflect these changes.

Fixes #8, fixes #10

### DIFF
--- a/cdn_lambda/functions/map_to_s3/map_to_s3.py
+++ b/cdn_lambda/functions/map_to_s3/map_to_s3.py
@@ -34,33 +34,33 @@ class LambdaClient(object):
     def handler(self, event, context):
         # pylint: disable=unused-argument
 
-        # Get uri from origin request event
         request = event["Records"][0]["cf"]["request"]
-        web_uri = request["uri"]
 
         LOG.info(
             "Querying '%s' table for '%s'...",
             self.conf["table"]["name"],
-            web_uri,
+            request["uri"],
         )
 
-        iso_now = datetime.now(timezone.utc).isoformat(timespec="milliseconds")
-
-        print(event)
         query_result = self.db_client.query(
             TableName=self.conf["table"]["name"],
             Limit=1,
             ScanIndexForward=False,
             KeyConditionExpression="web_uri = :u and from_date <= :d",
             ExpressionAttributeValues={
-                ":u": {"S": web_uri},
-                ":d": {"S": str(iso_now)},
+                ":u": {"S": request["uri"]},
+                ":d": {
+                    "S": str(
+                        datetime.now(timezone.utc).isoformat(
+                            timespec="milliseconds"
+                        )
+                    )
+                },
             },
         )
-        print(query_result)
 
         if query_result["Items"]:
-            LOG.info("Item found for '%s'", web_uri)
+            LOG.info("Item found for '%s'", request["uri"])
 
             try:
                 # Update request uri to point to S3 object key
@@ -77,7 +77,7 @@ class LambdaClient(object):
 
                 raise err
         else:
-            LOG.info("No item for '%s'", web_uri)
+            LOG.info("No item found for '%s'", request["uri"])
 
             # Report 404 to prevent attempts on S3
             return {
--- a/cdn_lambda/functions/map_to_s3/map_to_s3.py
+++ b/cdn_lambda/functions/map_to_s3/map_to_s3.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import os
 from datetime import datetime, timezone
 
 import boto3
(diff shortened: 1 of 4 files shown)


nathanegillett:
@rohanpm can we add "feedok" (Alexei) to the team and as a reviewer?

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -1,38 +1,42 @@
...
 
-def lambda_handler(event, context):
+def load_conf(conf_file):
+    with open(conf_file, "r") as json_file:
+        content = json.load(json_file)
+
+    return content
+
+
+def lambda_handler(event, context, conf_file=None):
```
nathanegillett: conf_file parameter is really only for testing.
The json file name needs to be hard-coded here to work in AWS but auto-tests can't locate it.
This seems the only reasonable way to enable testing without mocking open().

docs/functions/map_to_s3.rst
```diff
@@ -14,24 +14,22 @@ Deployment
...
 The event for this function must be a CloudFront distribution origin-request to
 an S3 bucket.
 
-Environment Variables
-^^^^^^^^^^^^^^^^^^^^^
-- DB_TABLE_NAME
-    (Required)
+Configuration
+^^^^^^^^^^^^^
+The map_to_s3 function must be deployed with the map_to_s3.json configuration
```
rohanpm: What does that mean exactly?  The file must be within the deployment .zip ?

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -1,32 +1,36 @@
...
     # pylint: disable=unused-argument
 
-    # This AWS Lambda function must be deployed with DB_TABLE_NAME and
-    # DB_TABLE_REGION environment variables
-    table_name = os.getenv("DB_TABLE_NAME")
-    table_region = os.getenv("DB_TABLE_REGION")
+    conf_file = conf_file or "map_to_s3.json"
+    conf = load_conf(conf_file)
```
rohanpm: I don't think we should be doing this here.  Loading a JSON file is not all that expensive, but nevertheless doing it for every single request is unnecessarily wasteful and it'd be best to figure out how to handle one-time initialization like this.  (This is https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html point "Take advantage of execution context reuse to improve the performance of your function")

I think we should be doing something more like this:

```python
class PreparedLambda:
  def __init__(self, conf_file="map_to_s3.json"):
    # This code runs only when the package is initially imported
    self.conf = load_conf(conf_file)

  def lambda_handler(self, event, context):
    # ...lambda_handler body, as it is now; may refer to self.conf as needed

# Then make a 'lambda_handler' available at the module level which is bound to
# a particular instance of the class
lambda_handler = PreparedLambda().lambda_handler
```

That would result in the config file being loaded only once, when map_to_s3 is imported.  It doesn't compromise testability of the code, since tests can freely construct a PreparedLambda pointing at the desired config.

docs/functions/map_to_s3.rst
nathanegillett: Yes

cdn_lambda/functions/map_to_s3/map_to_s3.py
nathanegillett: I like this. Going to take one step further and initialize the boto3 client in the class as well.

docs/functions/map_to_s3.rst
rohanpm: No, it was just for me to understand during review.  I think this is OK - creation of deployment package obviously should be automated at some point, then I wouldn't see much point duplicating all the details in the docs.

nathanegillett:
Will go ahead with merge, as new changes are minor.

PR #13, opened 2020-02-27
Make deployment cleaner

rohanpm:
- Add a tox configuration for creating a deployment zip file, so there's a
  standard way of doing it
- Add top-level export for lambda function, so that internal module structure
  doesn't influence AWS configs

See individual commits for more details.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,27 @@ deps=
 commands=pytest {posargs}
 whitelist_externals=sh
 
+# Create a lambda deployment package.
+[testenv:package]
+skip_install=true
+deps=
+whitelist_externals=
+    sh
+    rm
+    cp
+commands=
+    rm -rf ./package
+
+    # --no-deps to avoid using anything untrusted from PyPI;
+    # currently all deps are expected to be available from lambda runtime anyway.
+    pip install --no-deps --target ./package .
+
+    # Always using hardcoded config for now.
+    cp cdn_lambda/functions/map_to_s3/map_to_s3.json package
+
+    sh -c 'cd package && zip -r ../package.zip .'
+    rm -rf ./package
+
 [testenv:static]
 deps=
 	-rtest-requirements.txt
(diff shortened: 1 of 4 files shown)


rohanpm:
@crungehottman @nathanegillett 

The background here is that I was looking at the lambda functions which had been previously deployed, and saw for instance a 'map_to_s3.lambda_handler' which seems to require zipping up specifically the cdn_lambda/functions/map_to_s3 directory.  It seems rather messy, e.g. it breaks normal Python import behavior (imagine if map_to_s3 imports some cdn_lambda.utils - it'll work fine for autotests in this repo, then break when you go to deploy it).

tox.ini
```diff
@@ -8,6 +8,27 @@ deps=
...
+    rm -rf ./package
+
+    # --no-deps to avoid using anything untrusted from PyPI;
+    # currently all deps are expected to be available from lambda runtime anyway.
+    pip install --no-deps --target ./package .
+
+    # Always using hardcoded config for now.
+    cp cdn_lambda/functions/map_to_s3/map_to_s3.json package
+
+    sh -c 'cd package && zip -r ../package.zip .'
```
nathanegillett: I don't think this is deployable, as you only seem to include the configuration file in package.zip.
I think you'd need to copy the whole directory, like;
```suggestion
    # Always using hardcoded config for now.
    cp -R cdn_lambda/functions/map_to_s3 package

    sh -c 'cd package && zip -r ../package.zip .'
```

rohanpm: > I don't think this is deployable, as you only seem to include the configuration file in package.zip.

It is deployable, I've got it deployed now.  It contains both the code and config.

The "pip install" command above installs the code, using the standard Python install procedure - that's part of what this commit is about (rather than having to know which directory is supposed to be copied).

So, what happens is:

- "pip install" means that "package" directory ends up with the entire cdn_lambda/* python egg
- Then cp of the .json file puts the config into place as well.  The code loads the .json file from the current working directory; since loading the config works correctly, I guess that the Lambda runtime sets the working directory to the root of the deployment package.

I thought about making pip install the config file as well since setuptools does support that, but decided against it because it seems like this method of managing the config file is temporary.

PR #14, opened 2020-03-16
Manage and bundle exact dependency versions

rohanpm:
Previously, the package produced by "tox -e package" did not
bundle any dependencies.  Although this worked, let's start
bundling dependencies instead and managing their versions,
for a couple of reasons:

- It turns out this is actually the recommended practice even
  for deps such as boto3, despite the lambda runtime providing
  a copy of it; although there is much confusion on this point.
  See [1] for a discussion about this.

- If we don't bundle dependencies, then we have no way to roll
  back changes to those dependencies. This puts us at risk if
  the lambda runtime updates a dependency in a manner such that
  our code stops working.

- We'll likely want to introduce dependencies on some other
  packages owned by ourselves at some point, so let's figure out
  how to do that securely. Declaring dependencies with --hash
  and letting dependabot update them seems to achieve this, and
  dependabot supports updating those automatically while
  maintaining hashes.

For that last point, the version of boto3 included here was
deliberately chosen as a slightly older release, so that we can
try having dependabot update the dependency.

[1] https://www.serverlessops.io/blog/aws-lambda-and-python-boto3-bundling


tox.ini
```diff
@@ -16,11 +18,12 @@ whitelist_externals=
     sh
     rm
     cp
+commands_pre=
```
nathanegillett: Is this doing anything without a command?

rohanpm: It's just overriding the commands_pre from the [testenv] section, setting it back to empty.

[testenv] is defining the default commands_pre for all test environments (including "dynamic" environments which are not listed explicitly, like py37, py38 etc), then these which don't need it are resetting it back to empty.

PR #15, opened 2020-03-18
Provide origin path aliases for map_to_s3 [DELIVERY-9004]

nathanegillett:
This commit adds path alias mapping in map_to_s3 configuration file and
a method in the function itself to transform URIs. If the beginning of a
URI matches a key specified in the configuration file, it is replaced by
the key's value.

### DIFF
--- a/cdn_lambda/functions/map_to_s3/map_to_s3.json
+++ b/cdn_lambda/functions/map_to_s3/map_to_s3.json
@@ -2,5 +2,9 @@
   "table": {
     "name": "exodus",
     "region": "us-east-1"
-  }
+  },
+  "uri_aliases": [
+    ["/content/origin/", "/origin/"],
+    ["/origin/rpm/", "/origin/rpms/"]
+  ]
 }
--- a/cdn_lambda/functions/map_to_s3/map_to_s3.py
+++ b/cdn_lambda/functions/map_to_s3/map_to_s3.py
@@ -31,24 +31,29 @@ class LambdaClient(object):
 
         return self._db_client
 
+    def uri_alias(self, uri):
+        # NOTE: Aliases are processed in the order they are listed
+        for alias in self.conf["uri_aliases"]:
+            if uri.startswith(alias[0]):
+                uri = uri.replace(alias[0], alias[1])
+        return uri
+
     def handler(self, event, context):
         # pylint: disable=unused-argument
 
         request = event["Records"][0]["cf"]["request"]
+        uri = self.uri_alias(request["uri"])
+        table = self.conf["table"]["name"]
 
-        LOG.info(
-            "Querying '%s' table for '%s'...",
-            self.conf["table"]["name"],
-            request["uri"],
-        )
+        LOG.info("Querying '%s' table for '%s'...", table, uri)
 
         query_result = self.db_client.query(
-            TableName=self.conf["table"]["name"],
+            TableName=table,
             Limit=1,
             ScanIndexForward=False,
             KeyConditionExpression="web_uri = :u and from_date <= :d",
             ExpressionAttributeValues={
-                ":u": {"S": request["uri"]},
+                ":u": {"S": uri},
                 ":d": {
                     "S": str(
                         datetime.now(timezone.utc).isoformat(
@@ -60,7 +65,7 @@ class LambdaClient(object):
         )
 
         if query_result["Items"]:
-            LOG.info("Item found for '%s'", request["uri"])
+            LOG.info("Item found for '%s'", uri)
 
             try:
                 # Update request uri to point to S3 object key
@@ -77,7 +82,7 @@ class LambdaClient(object):
 
                 raise err
         else:
-            LOG.info("No item found for '%s'", request["uri"])
+            LOG.info("No item found for '%s'", uri)
 
             # Report 404 to prevent attempts on S3
             return {
@@ -87,4 +92,4 @@ class LambdaClient(object):
 
 
 # Make handler available at module level
-lambda_handler = LambdaClient().handler
+lambda_handler = LambdaClient().handler  # pylint: disable=invalid-name
--- a/tests/functions/test_map_to_s3.py
+++ b/tests/functions/test_map_to_s3.py
@@ -3,26 +3,58 @@ import pytest
 import mock
 import json
 
-TEST_PATH = "www.example.com/content/file.ext"
+TEST_PATH = "/origin/rpms/repo/ver/dir/filename.ext"
 MOCKED_DT = "2020-02-17T15:38:05.864+00:00"
 CONF_PATH = "cdn_lambda/functions/map_to_s3/map_to_s3.json"
 
 
+@pytest.mark.parametrize(
+    "req_uri, real_uri",
+    [
+        (
+            "/origin/rpm/repo/ver/dir/filename.ext",
+            "/origin/rpms/repo/ver/dir/filename.ext",
+        ),
+        (
+            "/content/origin/repo/ver/dir/filename.ext",
+            "/origin/repo/ver/dir/filename.ext",
+        ),
+        (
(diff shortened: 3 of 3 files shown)


cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -31,6 +31,12 @@ def db_client(self):
 
         return self._db_client
 
+    def uri_alias(self, uri):
+        for key, value in self.conf["uri_aliases"].items():
+            if key in uri:
```
rohanpm: It shouldn't be key "in" uri, it should be if the uri path startswith key, right?  The key could appear somewhere other than the beginning, in which case it'd be wrong to replace it.

It's not a negligible risk, for example there is a project associated with Red Hat called "origin" (openshift origin) so it's not that unlikely that strings such as /content/origin/ or /origin/rpm/ could appear somewhere in the URL other than at the beginning.

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -31,6 +31,12 @@ def db_client(self):
 
         return self._db_client
 
+    def uri_alias(self, uri):
+        for key, value in self.conf["uri_aliases"].items():
```
rohanpm: Question: are the items guaranteed to be processed here in the same order they were listed in the JSON file?  Because the code only seems to work correctly if that's the case.  (For instance /content/origin/rpm/foobar only works if /content/origin/ is processed and then /origin/rpm/ is processed.)

At the very least I would recommend adding a comment here pointing out that order matters here, as it's not at all obvious.  Maybe even going further and making the config file use a list instead of a dict.

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -48,7 +54,7 @@ def handler(self, event, context):
             ScanIndexForward=False,
             KeyConditionExpression="web_uri = :u and from_date <= :d",
             ExpressionAttributeValues={
-                ":u": {"S": request["uri"]},
+                ":u": {"S": self.uri_alias(request["uri"])},
```
rohanpm: The LOG.info a few lines above here is now lying (It will say "Querying <table_name> table for <uri>" where uri isn't what it's really querying for).  Can that please be fixed?

Ideally the logs contain enough info to judge whether the alias resolving is working correctly.

tests/functions/test_map_to_s3.py
```diff
@@ -8,6 +8,24 @@
 CONF_PATH = "cdn_lambda/functions/map_to_s3/map_to_s3.json"
 
 
+@pytest.mark.parametrize(
+    "uri",
+    [
+        "www.example.com/content/origin/rpms/repo/ver/filename",
+        "www.example.com/origin/rpm/repo/ver/filename",
+        "www.example.com/content/origin/rpm/repo/ver/filename",
```
rohanpm: Relating to that point about "key in" vs "startswith", it would be good to also add a case where the alias appears somewhere other than the beginning of a URL here.

tests/functions/test_map_to_s3.py
```diff
@@ -8,6 +8,24 @@
...
+    ],
+    ids=[
+        "content/origin -> origin",
+        "origin/rpm -> origin/rpms",
+        "content/origin/rpm -> origin/rpms",
+    ],
+)
+def test_map_to_s3_uri_alias(uri):
+    expected = "www.example.com/origin/rpms/repo/ver/filename"
+    assert LambdaClient(conf_file=CONF_PATH).uri_alias(uri) == expected
```
rohanpm: This test is OK but it doesn't really seem sufficient.  Wouldn't the test still pass even if you completely forgot to call uri_alias from within handler?  (Or if someone accidentally removed it during a refactor.)

In my opinion, there should be at least one test which covers this behavior from the top level of the handler.

cdn_lambda/functions/map_to_s3/map_to_s3.py
nathanegillett: This was intentional to represent the original request URI but I can change it.

cdn_lambda/functions/map_to_s3/map_to_s3.py
nathanegillett: Ah, I didn't think that was a realistic scenario. I'll make the change but it has consequences as you've already noticed.

cdn_lambda/functions/map_to_s3/map_to_s3.py
nathanegillett: It seems that, in Python3 at least, the order in the JSON file is maintained. I considered sorting the dict to ensure /content/ was always processed first but having the dict alphabetized may cause issues for other alias sets.

I'm not sure how the dict could be converted into a list or how that might help.

tests/functions/test_map_to_s3.py
nathanegillett: Done.

tests/functions/test_map_to_s3.py
nathanegillett: Done.

nathanegillett:
@rohanpm can we enable the "dismiss stale pull request approvals when new commits are pushed" branch protection feature?

cdn_lambda/functions/map_to_s3/map_to_s3.py
rohanpm: > I'm not sure how the dict could be converted into a list or how that might help.

We control both the config file format and the code, so you wouldn't convert it to a list, you would just make the config file use a list in the first place, e.g.

```
  "uri_aliases": [
    ["/content/origin/", "/origin/"],
    ["/origin/rpm/", "/origin/rpms/"]
  ]
```

It would help because lists always retain order, unlike dicts which generally have an undefined order.

Do we deploy on at least Python 3.7?  Per comments at https://docs.python.org/3/library/json.html (search for "preserve input and output order"), that seems to be the first version at which this behavior can be relied on, that was the first Python version where dicts became ordered by default.  tox.ini is only testing with Python 3.6.

tests/functions/test_map_to_s3.py
```diff
@@ -3,26 +3,47 @@
...
 import json
 
-TEST_PATH = "www.example.com/content/file.ext"
+TEST_PATH = "/origin/rpms/repo/ver/dir/filename.ext"
 MOCKED_DT = "2020-02-17T15:38:05.864+00:00"
 CONF_PATH = "cdn_lambda/functions/map_to_s3/map_to_s3.json"
 
 
+@pytest.mark.parametrize(
+    "req_uri, real_uri",
```
rohanpm: Can we please also have a case where we do NOT do any rewriting?  (e.g. to protect against a bug where matching no aliases at all causes a crash.)

cdn_lambda/functions/map_to_s3/map_to_s3.py
nathanegillett: Gotcha. I didn't know list order was always maintained. In that case, I'll switch to a 2D-list, as you suggest.

tests/functions/test_map_to_s3.py
nathanegillett: Done

cdn_lambda/functions/map_to_s3/map_to_s3.json
```diff
@@ -2,5 +2,9 @@
   "table": {
     "name": "exodus",
     "region": "us-east-1"
-  }
+  },
+  "uri_aliases": [
```
rohanpm: I just want to mention, in case it wasn't clear, that I plan to do a later change which removes the aliases from this repo and uses cdn-definitions package instead.

nathanegillett: Yeah, I'd  prefer that as well.

PR #16, opened 2020-03-18
Add origin_response lambda function

crungehottman:
The cache_control function appends a "Cache-Control: max_age=600" header to HTTP responses from the origin server.

Resolves: DELIVERY-8986

### DIFF
--- a/cdn_lambda/functions/map_to_s3/map_to_s3.py
+++ b/cdn_lambda/functions/map_to_s3/map_to_s3.py
@@ -8,7 +8,7 @@ LOG = logging.getLogger("map-to-s3-lambda")
 
 
 class LambdaClient(object):
-    def __init__(self, conf_file="map_to_s3.json"):
+    def __init__(self, conf_file="lambda_config.json"):
         self._conf_file = conf_file
 
         self._conf = None
@@ -68,6 +68,11 @@ class LambdaClient(object):
             LOG.info("Item found for '%s'", uri)
 
             try:
+                # Add custom header containing the original request uri
+                request["headers"]["exodus-original-uri"] = [
+                    {"key": "exodus-original-uri", "value": request["uri"]}
+                ]
+
                 # Update request uri to point to S3 object key
                 request["uri"] = (
                     "/" + query_result["Items"][0]["object_key"]["S"]
@@ -85,10 +90,7 @@ class LambdaClient(object):
             LOG.info("No item found for '%s'", uri)
 
             # Report 404 to prevent attempts on S3
-            return {
-                "status": "404",
-                "statusDescription": "Not Found",
-            }
+            return {"status": "404", "statusDescription": "Not Found"}
 
 
 # Make handler available at module level
--- a/cdn_lambda/functions/origin_response/origin_response.py
+++ b/cdn_lambda/functions/origin_response/origin_response.py
@@ -1,0 +1,63 @@
+import json
+import logging
+import re
+
+LOG = logging.getLogger("origin-response")
+
+
+class LambdaClient(object):
+    def __init__(self, conf_file="lambda_config.json"):
+        self._conf_file = conf_file
+        self._conf = None
+
+    @property
+    def conf(self):
+        if not self._conf:
+            with open(self._conf_file, "r") as json_file:
+                self._conf = json.load(json_file)
+
+        return self._conf
+
+    def handler(self, event, context):
+        # pylint: disable=unused-argument
+
+        request = event["Records"][0]["cf"]["request"]
+        response = event["Records"][0]["cf"]["response"]
+
+        max_age = self.conf["headers"]["max_age"]
+        max_age_pattern_whitelist = [
+            ".+/PULP_MANIFEST",
+            ".+/listing",
+            ".+/repodata/repomd.xml",
+            ".+/ostree/repo/refs/heads/.*/.*",
+        ]
+
+        try:
+            original_uri = request["headers"]["exodus-original-uri"][0][
+                "value"
+            ]
+
+        except (KeyError):
+            LOG.debug(
+                "Could not read exodus-original-uri from response", exc_info=1
+            )
+            original_uri = None
+
+        if original_uri is not None:
+            for pattern in max_age_pattern_whitelist:
+                if re.match(pattern, original_uri):
+                    response["headers"]["cache-control"] = [
+                        {
+                            "key": "Cache-Control",
+                            "value": "max-age={}".format(max_age),
+                        }
+                    ]
+                    LOG.info(
+                        "Cache-Control header added for '%s'", original_uri
+                    )
+
+        return response
+
+
+# Make handler available at module level
+lambda_handler = LambdaClient().handler
--- a/tests/functions/test_origin_response.py
+++ b/tests/functions/test_origin_response.py
@@ -1,0 +1,68 @@
+from cdn_lambda.functions.origin_response.origin_response import LambdaClient
+import json
+import pytest
+
+CONF_PATH = "cdn_lambda/functions/lambda_config.json"
+
+# Load max_age from conf file:
+with open(CONF_PATH, "r") as json_file:
+    conf = json.load(json_file)
+max_age = conf["headers"]["max_age"]
+
+
+@pytest.mark.parametrize(
+    "test_input",
+    [
+        "/some/repo/listing",
+        "/some/repo/repodata/repomd.xml",
+        "/some/repo/ostree/repo/refs/heads/ok/test",
+    ],
+)
+def test_origin_response_valid_headers(
+    test_input,
+    expected=[{"key": "Cache-Control", "value": "max-age={}".format(max_age)}],
+):
+    event = {
+        "Records": [
+            {
+                "cf": {
+                    "request": {
+                        "headers": {
+                            "exodus-original-uri": [
+                                {
+                                    "key": "exodus-original-uri",
+                                    "value": test_input,
+                                }
+                            ]
+                        }
+                    },
+                    "response": {"headers": {}},
+                }
+            }
+        ]
+    }
+
+    response = LambdaClient(conf_file=CONF_PATH).handler(event, context=None)
+    assert response["headers"]["cache-control"] == expected
+
+
+@pytest.mark.parametrize(
+    "test_input",
+    ["/some/repo/some-rpm.rpm", "/some/repo/ostree/repo/refs/heads/nope"],
+)
+def test_origin_response_empty_headers(test_input, expected={}):
+    event = {
+        "Records": [
+            {"cf": {"request": {"headers": {}}, "response": {"headers": {}}}}
+        ]
+    }
+
+    response = LambdaClient(conf_file=CONF_PATH).handler(event, context=None)
+    assert response["headers"] == expected
+
+
+def test_origin_response_missing_headers():
+    event = {"Records": [{"cf": {"request": {}, "response": {}}}]}
+
+    response = LambdaClient(conf_file=CONF_PATH).handler(event, context=None)
+    assert response == {}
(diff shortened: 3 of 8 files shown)


rohanpm:
I don't think it should be named cache_control, should it?  Because we expect that we will need to add some more behaviors onto origin-response events, and we'll need all of that to be handled by one lambda function, so we would end up with a function named cache_control which is actually doing a bunch more.

Actually, the internal name doesn't matter much, but could we please have an entry for the function in ``cdn_lambda/__init__.py`` named ``origin_response`` to match the existing ``origin_request``?

nathanegillett:
@rohanpm, @crungehottman, I wouldn't be opposed to naming the function "origin_response" and changing "map_to_s3" to "origin_request". I don't foresee swapping out request or response functions on our distributions.

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -80,10 +85,7 @@ def handler(self, event, context):
             LOG.info("No item found for '%s'", request["uri"])
 
             # Report 404 to prevent attempts on S3
-            return {
-                "status": "404",
-                "statusDescription": "Not Found",
-            }
+            return {"status": "404", "statusDescription": "Not Found"}
```
crungehottman: note: black reformatted this line

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -63,6 +63,11 @@ def handler(self, event, context):
             LOG.info("Item found for '%s'", request["uri"])
 
             try:
+                # Add custom header containing the original request uri
+                request["headers"]["original-uri"] = [
```
rohanpm: Just confirming, have you already deployed this in some environment and seen it working?

Also, could we please use something which stands out more for our own custom headers; how about a standard prefix, like "exodus-" ?

At a glance, if someone is looking at code referencing an "original-uri" header, it won't be obvious that this is something we invented, while "exodus-original-uri" makes it clear.  Also it can't be ruled out that a future version of the CloudFront platform would introduce its own original-uri header, as the name is fairly generic.

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,29 @@
...
+    request = event["Records"][0]["cf"]["request"]
+    original_uri = request["headers"]["original-uri"][0]["value"]
+
+    response = event["Records"][0]["cf"]["response"]
+
+    max_age_pattern_whitelist = [
+        ".+/PULP_MANIFEST",
+        ".+/listing",
+        ".+/repodata/repomd.xml",
+        ".+/ostree/repo/refs/heads/.+/.+",
```
rohanpm: Those last two .+ should probably be [^/]+, right?  Else it actually matches any number of / >= 1, but if that's acceptable then we may as well just have written .*.

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,29 @@
...
+import re
+
+LOG = logging.getLogger("origin-response")
+
+
+def lambda_handler(event, context):
+    # pylint: disable=unused-argument
+
+    request = event["Records"][0]["cf"]["request"]
+    original_uri = request["headers"]["original-uri"][0]["value"]
```
rohanpm: Could this please be written in such a way that it won't crash if the header is absent?  It would be nice to keep the lambda able to execute without crashing when passed a "standard" CloudFront event, and not have it 100% rely on custom fields we've added.

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,29 @@
...
+
+    max_age_pattern_whitelist = [
+        ".+/PULP_MANIFEST",
+        ".+/listing",
+        ".+/repodata/repomd.xml",
+        ".+/ostree/repo/refs/heads/.+/.+",
+    ]
+
+    for pattern in max_age_pattern_whitelist:
+        if re.search(pattern, original_uri):
```
rohanpm: It seems that either the patterns should end with $ or this should be an re.match, to ensure the pattern matches all the way up to the end.

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,29 @@
...
+            response["headers"]["cache-control"] = [
+                {"key": "Cache-Control", "value": "max-age=600"}
```
rohanpm: I think the issue description asked for the number to be configurable.  Could that please be arranged?  e.g. the number loaded from the config file.

If so, the config file probably should no longer be named map_to_s3.json too.

tests/functions/test_origin_response.py
```diff
@@ -0,0 +1,118 @@
...
+    event = {
+        "Records": [
+            {
+                "cf": {
+                    "request": {
+                        "headers": {
+                            "original-uri": [
+                                {
+                                    "key": "original-uri",
+                                    "value": "www.example.com/some/repo/some-rpm.rpm",
```
rohanpm: I don't think this is accurate testdata. I had to look this up while reviewing another pending PR and I found that uri does not include the host portion, and always starts with "/".  Could this please be accurate too?

The main reason to care about it is to avoid misleading readers of this code, i.e. something looking at this test and trying to understand the code would be led to wrongly guess that original-uri includes hostname, when it actually won't (right?)

I know that the other test already has this mistake, but let's try to avoid that misunderstanding propagating any further.

tests/functions/test_origin_response.py
```diff
@@ -0,0 +1,118 @@
...
+    event = {
+        "Records": [
+            {
+                "cf": {
+                    "request": {
+                        "headers": {
+                            "original-uri": [
+                                {
+                                    "key": "original-uri",
+                                    "value": "www.example.com/some/repo/listing",
```
rohanpm: Same here about whether or not host is in uri.

tests/functions/test_origin_response.py
```diff
@@ -0,0 +1,118 @@
...
+
+    response = lambda_handler(event, context=None)
+
+    # listing files should contain cache headers
+    assert response["headers"]["cache-control"] == [
+        {"key": "Cache-Control", "value": "max-age=600"}
+    ]
+
+
+def test_origin_response_repomd():
```
rohanpm: You may want to look into pytest.mark.parametrize, it allows you to run the same test code multiple times with different inputs and expected values.  Right now this seems to have the same test duplicated a few times and only with different URI values.  A parametrized test fits well here.

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if original_uri is not None:
+            for pattern in max_age_pattern_whitelist:
+                if re.match(pattern, original_uri):
+                    response["headers"]["cache-control"] = [
+                        {
+                            "key": "Cache-Control",
+                            "value": "max-age={}".format(max_age),
+                        }
+                    ]
```
nathanegillett: Okay, maybe I'm missing something about structure for headers. Are the "key" and "value" keys necessary?

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -63,6 +63,11 @@ def handler(self, event, context):
             LOG.info("Item found for '%s'", request["uri"])
 
             try:
+                # Add custom header containing the original request uri
+                request["headers"]["exodus-original-uri"] = [
+                    {"key": "exodus-original-uri", "value": request["uri"]}
+                ]
+
```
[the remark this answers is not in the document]
crungehottman: I attempted this, but received a `LambdaValidationError from cloudfront` error, which seems to be related to a discrepancy between `"key"`'s value and the header's key: https://forums.aws.amazon.com/message.jspa?messageID=796827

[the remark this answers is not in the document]
nathanegillett: Oh, I see, this is some Lambda@Edge specific requirement. Thanks!

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,63 @@
...
+
+        request = event["Records"][0]["cf"]["request"]
+        response = event["Records"][0]["cf"]["response"]
+
+        max_age = self.conf["headers"]["max_age"]
+        max_age_pattern_whitelist = [
+            ".+/PULP_MANIFEST",
+            ".+/listing",
+            ".+/repodata/repomd.xml",
+            ".+/ostree/repo/refs/heads/.*/.*",
```
crungehottman: I'm a regex newbie -- I don't think this pattern is optimal, but it works. Feel free to suggest a more optimal regex.

cdn_lambda/functions/map_to_s3/map_to_s3.py
crungehottman: > Just confirming, have you already deployed this in some environment and seen it working?

Yes. I deployed this in my 'caleigh' cloudfront distribution with the 'origin_reponse' lambda function.

cdn_lambda/functions/origin_response/origin_response.py
crungehottman: regex was updated in latest commit, but I'm still not sure if it's correct

cdn_lambda/functions/map_to_s3/map_to_s3.py
```diff
@@ -8,7 +8,7 @@
 
 
 class LambdaClient(object):
-    def __init__(self, conf_file="map_to_s3.json"):
+    def __init__(self, conf_file="lambda_conf.json"):
```
rohanpm: This is slightly different from the filename used in the repo, is that intentional?

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,63 @@
...
+        ]
+
+        try:
+            original_uri = request["headers"]["exodus-original-uri"][0][
+                "value"
+            ]
+
+        except (KeyError):
+            LOG.debug(
+                "Could not add Cache-Control header 'exodus-original-uri'"
```
rohanpm: This log message seems to be a bit mangled, it reads like 'exodus-original-uri' is a cache-control header, and that we failed to add that header, both of which are not accurate.

Also it should probably have exc_info so that the exception isn't lost entirely, because actually you don't know for certain that the KeyError was raised due to missing exodus-original-uri.

Would something like this be more appropriate?

```suggestion
                "Could not read exodus-original-uri from response", exc_info=1
```

exc_info=1 means it will log backtrace of the exception currently being handled.

cdn_lambda/functions/map_to_s3/map_to_s3.py
crungehottman: Nice catch! That was unintentional. Fixed.

crungehottman:
> I think this will break "tox -e package" because the config file path is written there (in tox.ini), could you please have a look?

It did break `tox - e package`. I updated the path to the config file in `tox.ini` and now `tox -e package` is passing.

> I'm not sure how you generate your deployment packages right now during development but you should maybe look into using that command if you aren't already, ideally we are all using the same method to make a deployment package so we will find the same issues.

I have not been generating deployment packages, but I will be sure to include `tox -e package` in my development workflow.

> CI also really ought to be testing that tox command, I'll make a separate PR for that.

Thanks!

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -0,0 +1,63 @@
+import json
+import logging
+import re
+
+LOG = logging.getLogger("origin-response")
+
+
+class LambdaClient(object):
+    def __init__(self, conf_file="lambda_conf.json"):
```
rohanpm: This seems broken:

- origin_request uses lambda_config.json
- origin_response uses lambda_conf.json

So you would have to deploy two copies of the config file under different names for this to work - doesn't seem intentional.

rohanpm:
Resolved the conflicts.  I'll submit this for you now so it lands prior to repo rename.

PR #29, opened 2020-03-26
docs: add Architecture page with diagram and summary of components

rohanpm:
Adds a new page to the online docs which contain an overview of
how all the components fit together, plus a description.
It is hoped to serve as a useful entry point for someone trying
to learn what this project is about.

### DIFF
--- a/docs/arch.gv
+++ b/docs/arch.gv
@@ -1,0 +1,139 @@
+digraph {
+    ranksep="1.4";
+
+    # These are arranged and labelled to communicate the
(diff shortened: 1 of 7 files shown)


docs/arch.gv
```diff
@@ -0,0 +1,139 @@
+digraph {
```
rohanpm: A rendered form of this graph can be seen here:

https://rohanpm.github.io/cdn-lambda/arch.html

PR #32, opened 2020-03-27
Rename project from cdn-lambda to exodus-lambda

rohanpm:
cdn-lambda is too generic, as it could be understood as describing the technology in use
(lambda functions within CloudFront CDN).  Name it exodus-lambda to associate it more
strongly with Exodus CDN.

This pull request is mostly a straight rename, with a couple of tiny exceptions:

- link to PyPI was removed, as we haven't and probably won't release this to PyPI
- replaced _ with - in the egg name from setup.py, as there is no need to avoid - here, and by convention the egg name usually matches exactly the repo name.


rohanpm:
Coveralls status update didn't arrive for some reason, I am thinking maybe due to the repo rename.  I found that the renamed repo initially wasn't at https://coveralls.io/repos/new (I added it now).  Please ignore the missing coveralls result.

nathanegillett:
I'm going to go ahead with this merge, @rohanpm. @crungehottman and I can then update our forks and make any necessary adjustments.

rohanpm:
We still seem to be having problems with coveralls after repo rename... I've disabled it from the required checks, hopefully temporarily.

PR #34, opened 2020-03-27
Support want-digest request headers [DELIVERY-8988]

nathanegillett:
This commit enables id-sha-256 digest header in the response for any
want-digest, regardless of requested algorithm.

### DIFF
--- a/tests/functions/test_origin_response.py
+++ b/tests/functions/test_origin_response.py
@@ -13,30 +13,29 @@ max_age = conf["headers"]["max_age"]
 
 
 @pytest.mark.parametrize(
-    "test_input",
+    "original_uri, want_digest",
     [
-        "/some/repo/listing",
-        "/some/repo/repodata/repomd.xml",
-        "/some/repo/ostree/repo/refs/heads/ok/test",
+        ("/some/repo/listing", True),
+        ("/some/repo/repodata/repomd.xml", True),
+        ("/some/repo/ostree/repo/refs/heads/ok/test", False),
     ],
 )
-def test_origin_response_valid_headers(
-    test_input,
-    expected=[{"key": "Cache-Control", "value": "max-age={}".format(max_age)}],
-):
+def test_origin_response_valid_headers(original_uri, want_digest):
     event = {
         "Records": [
             {
                 "cf": {
                     "request": {
+                        "uri": "/be7f3007df3e51fb48fff57da9c01c52e6b8e60eceacab"
+                        "7aaf0e05b57578493a",
                         "headers": {
                             "exodus-original-uri": [
                                 {
                                     "key": "exodus-original-uri",
-                                    "value": test_input,
+                                    "value": original_uri,
                                 }
                             ]
-                        }
+                        },
                     },
                     "response": {"headers": {}},
                 }
@@ -44,15 +43,33 @@ def test_origin_response_valid_headers(
         ]
     }
 
+    expected_headers = {
+        "cache-control": [
+            {"key": "Cache-Control", "value": f"max-age={max_age}"}
+        ]
+    }
+
+    if want_digest:
+        event["Records"][0]["cf"]["request"]["headers"]["want-digest"] = [
+            {"key": "Want-Digest", "value": "sha-256"}
+        ]
+        expected_headers["digest"] = [
+            {
+                "key": "Digest",
+                "value": "id-sha-256=vn8wB98+UftI//V9qcAcUua45g7OrKt6rw"
+                "4FtXV4STo=",
+            }
+        ]
+
     response = OriginResponse(conf_file=CONF_PATH).handler(event, context=None)
-    assert response["headers"]["cache-control"] == expected
+    assert response["headers"] == expected_headers
 
 
 @pytest.mark.parametrize(
     "test_input",
     ["/some/repo/some-rpm.rpm", "/some/repo/ostree/repo/refs/heads/nope"],
 )
-def test_origin_response_empty_headers(test_input, expected={}):
+def test_origin_response_empty_headers(test_input):
     event = {
         "Records": [
             {"cf": {"request": {"headers": {}}, "response": {"headers": {}}}}
@@ -60,7 +77,7 @@ def test_origin_response_empty_headers(test_input, expected={}):
     }
 
     response = OriginResponse(conf_file=CONF_PATH).handler(event, context=None)
-    assert response["headers"] == expected
+    assert response["headers"] == {}
 
 
 def test_origin_response_missing_headers():
--- a/exodus_lambda/functions/origin_response.py
+++ b/exodus_lambda/functions/origin_response.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from base64 import b64encode
 
 from .base import LambdaBase
 
@@ -12,6 +13,13 @@ class OriginResponse(LambdaBase):
 
         request = event["Records"][0]["cf"]["request"]
         response = event["Records"][0]["cf"]["response"]
+
+        if "headers" in request and "want-digest" in request["headers"]:
+            sum_hex = request["uri"].replace("/", "", 1)
+            sum_b64 = b64encode(bytes.fromhex(sum_hex)).decode()
+            response["headers"]["digest"] = [
+                {"key": "Digest", "value": f"id-sha-256={sum_b64}"}
+            ]
 
         max_age = self.conf["headers"]["max_age"]
         max_age_pattern_whitelist = [


cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +25,22 @@ def handler(self, event, context):
...
 
+        try:
+            want_digest = request["headers"]["want-digest"][0]["value"]
+        except KeyError:
+            LOG.debug(
+                "Could not read want-digest from response", exc_info=True
+            )
+            want_digest = None
+
+        if want_digest == "sha-256":
```
rohanpm: I think this is not the right strategy for matching want_digest header... the caller is allowed to request multiple digest types (e.g. if they support multiple) and this would break when the caller does that.

There was an example on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Want-Digest

> Want-Digest: SHA-512;q=0.3, sha-256;q=1, md5;q=0

I don't think we should silently ignore the Want-Digest request in that case.

Actually, it is legal for the server to respond with *any* digest type if the client asked for an unsupported type.  I think that means we don't need to look at the content of the header at all.  We can simply do "if the caller provided any Want-Digest header, give sha-256 digest".

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +25,22 @@ def handler(self, event, context):
...
+            response["headers"]["digest"] = [
+                {
+                    "key": "Digest",
+                    "value": f"sha-256={basename(request['uri'])}",
```
rohanpm: This isn't correct, the digest is supposed to be base64 encoded.

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +25,22 @@ def handler(self, event, context):
...
+        try:
+            want_digest = request["headers"]["want-digest"][0]["value"]
+        except KeyError:
+            LOG.debug(
+                "Could not read want-digest from response", exc_info=True
+            )
+            want_digest = None
+
+        if want_digest == "sha-256":
+            response["headers"]["digest"] = [
```
rohanpm: Question: do you also have to check whether the response from origin was successful?  If we got an error or other unusual response, I don't think we should add the digest.  This is a question since I don't know whether the origin-response event is designed so that the lambda is invoked also on failures (I guess that it probably is, since there's valid use-cases to do things with failures from origin).

cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +25,22 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         response = event["Records"][0]["cf"]["response"]
 
+        try:
+            want_digest = request["headers"]["want-digest"][0]["value"]
+        except KeyError:
+            LOG.debug(
+                "Could not read want-digest from response", exc_info=True
```
rohanpm: Request, not response.

```suggestion
                "Could not read want-digest from request", exc_info=True
```

tests/functions/test_origin_response.py
```diff
@@ -11,54 +11,66 @@
...
-    test_input,
-    expected=[{"key": "Cache-Control", "value": "max-age={}".format(max_age)}],
-):
+def test_origin_response_valid_headers(original_uri, want_digest):
     event = {
         "Records": [
             {
                 "cf": {
                     "request": {
+                        "uri": "/some/repo/a4ef27d",
```
rohanpm: Could the testdata please be more realistic?

1. In reality, there won't be a /some/repo prefix.  Could it please be removed, as this is misleading readers as to how the setup works.

2. Could we please use a full checksum rather than abbreviating?  I am worried abbreviating in the testdata is setting ourselves up for extra work later, e.g. if we introduce a feature which more strongly relies on the property that the keys are valid sha256 sums.

nathanegillett: Oops! Yeah, not sure what I was thinking with the path prefix. I'll replace it with a real sha256.

exodus_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +26,22 @@ def handler(self, event, context):
...
+        try:
+            want_digest = request["headers"]["want-digest"][0]["value"]
+        except KeyError:
+            LOG.debug("Could not read want-digest from request", exc_info=True)
+            want_digest = None
+
+        # TODO: Investigate cf-s3 response errors, ensure no digest if error
+        if want_digest:
+            sha256_object_key = basename(request["uri"])
+            response["headers"]["digest"] = [
```
nathanegillett: > Question: do you also have to check whether the response from origin was successful? If we got an error or other unusual response, I don't think we should add the digest. This is a question since I don't know whether the origin-response event is designed so that the lambda is invoked also on failures (I guess that it probably is, since there's valid use-cases to do things with failures from origin).

exodus_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +26,16 @@ def handler(self, event, context):
...
 
+        # TODO: Investigate cf-s3 response errors, ensure no digest if error
+        if "want-digest" in request["headers"]:
+            sha256_object_key = basename(request["uri"])
+            response["headers"]["digest"] = [
+                {
+                    "key": "Digest",
+                    "value": f"sha-256={b64encode(sha256_object_key)}",
+                }
+            ]
```
nathanegillett: > Question: do you also have to check whether the response from origin was successful? If we got an error or other unusual response, I don't think we should add the digest. This is a question since I don't know whether the origin-response event is designed so that the lambda is invoked also on failures (I guess that it probably is, since there's valid use-cases to do things with failures from origin).

cdn_lambda/functions/origin_response/origin_response.py
nathanegillett: I'm not sure what's meant by this. I've tried encoding the sha-256 object key but then it can't be used in the string. I would prefer to calculate the "real" checksum here but I'm not sure that's reasonable.

cdn_lambda/functions/origin_response/origin_response.py
nathanegillett: I'm only able to get any of the custom response headers, including this one, to appear on successful requests;
`HTTP/2 200 
content-type: binary/octet-stream
content-length: 2475
date: Mon, 30 Mar 2020 20:20:12 GMT
last-modified: Wed, 19 Feb 2020 22:29:14 GMT
etag: "038ce56c7a07d5f070572f1bf19fb939"
accept-ranges: bytes
server: AmazonS3
digest: sha-256=f8fc113a93a12fe6e2713cf550cd7d015d8defbdcb6fa00f9b85863b18c3efe4
x-cache: Miss from cloudfront
via: 1.1 11e42b3facdb2abb175876ba0e4b97f9.cloudfront.net (CloudFront)
x-amz-cf-pop: IAD79-C3
x-amz-cf-id: 922kbbsaKhGlQheZrlse-mnqDxwtdage4W3b6VfGieL2LokF65Fj3Q==`

Creating a table item which points to an invalid object key got me the following;

`HTTP/2 404 
content-length: 0
server: CloudFront
date: Mon, 30 Mar 2020 19:54:24 GMT
x-cache: Error from cloudfront
via: 1.1 15cde442051269a0307a638d23683c8c.cloudfront.net (CloudFront)
x-amz-cf-pop: IAD79-C3
x-amz-cf-id: nNilA3dXMxW07j2nL6IX7rpNUsaD9rsOQoCiwawNEMPaD3kLr46UkA==
age: 185`

If we want to be sure, we could check content-length header.

exodus_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +24,14 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         response = event["Records"][0]["cf"]["response"]
 
+        if "headers" in request and "want-digest" in request["headers"]:
+            response["headers"]["digest"] = [
+                {
+                    "key": "Digest",
+                    "value": f"sha-256={request['uri']}".replace("/", "", 1),
```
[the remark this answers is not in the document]
rohanpm: One more thing: after re-reading the spec, I think the correct digest for us to use is actually id-sha-256.  This explicitly means "digest of the content without encoding", which is always what we've stored in S3, so it makes sense.

sha-256 instead can be affected by encoding, so that could possibly be wrong if there are certain features enabled on the CDN like transparent compression (Content-Encoding: gzip or similar).

The only change is that sha-256= should be instead id-sha-256=.

exodus_lambda/functions/origin_response/origin_response.py
```diff
@@ -24,6 +24,12 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         response = event["Records"][0]["cf"]["response"]
 
+        if "headers" in request and "want-digest" in request["headers"]:
+            sha256_object_key = request["uri"].replace("/", "", 1)
+            response["headers"]["digest"] = [
+                {"key": "Digest", "value": f"id-sha-256={sha256_object_key}"}
```
rohanpm: This still needs the base64 encoding previously discussed.

nathanegillett: I thought id-sha-256 was digest of the content without encoding

rohanpm: It's a different encoding being discussed here.

Content-Encoding header can be used in HTTP to encode responses in various ways, e.g. "Content-Encoding: gzip" to indicate that the server has compressed the content and expects the client to decompress.

In that scenario, my understanding is that giving a digest of sha-256 means "This is the SHA256 of the encoded (compressed) data", while giving a digest of id-sha-256 means "This is the SHA256 of the original/decoded data".  The digest header itself still uses base64 encoding, the difference is what data has been checksummed.

nathanegillett: Okay, I think I understand.

PR #35, opened 2020-03-27
Minor fixes for origin_response

nathanegillett:
Corrects minor except and logging.Logger.debug syntax issues.
Changes string interpolation to f-string.
Disables pylint complaint for lambda_handler naming.

### DIFF
--- a/exodus_lambda/functions/origin_response/origin_response.py
+++ b/exodus_lambda/functions/origin_response/origin_response.py
@@ -37,19 +37,20 @@ class LambdaClient(object):
                 "value"
             ]
 
-        except (KeyError):
+        except KeyError:
             LOG.debug(
-                "Could not read exodus-original-uri from response", exc_info=1
+                "Could not read exodus-original-uri from response",
+                exc_info=True,
             )
             original_uri = None
 
-        if original_uri is not None:
+        if original_uri:
             for pattern in max_age_pattern_whitelist:
                 if re.match(pattern, original_uri):
                     response["headers"]["cache-control"] = [
                         {
                             "key": "Cache-Control",
-                            "value": "max-age={}".format(max_age),
+                            "value": f"max-age={max_age}",
                         }
                     ]
                     LOG.info(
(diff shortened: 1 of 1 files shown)


cdn_lambda/functions/origin_response/origin_response.py
```diff
@@ -37,23 +37,24 @@ def handler(self, event, context):
...
                     response["headers"]["cache-control"] = [
                         {
                             "key": "Cache-Control",
-                            "value": "max-age={}".format(max_age),
+                            "value": f"max-age={max_age}",
                         }
                     ]
                     LOG.info(
-                        "Cache-Control header added for '%s'", original_uri
+                        f"Cache-Control header added for '{original_uri}'"
```
rohanpm: Pylint complains about this and I agree, we should still let the logger take care of interpolation.  f-strings doesn't seem to change this.

This is the reason for CI failure.

exodus_lambda/functions/origin_response/origin_response.py
```diff
@@ -60,4 +61,4 @@ def handler(self, event, context):
 
 
 # Make handler available at module level
-lambda_handler = LambdaClient().handler
+lambda_handler = LambdaClient().handler  # pylint: disable=invalid-name
```
rohanpm: I know this isn't the first time this has been added... but I don't really understand why this is needed.  pylint shouldn't have a problem determining that an instance of LambdaClient has a 'handler' property.  pylint as run by Travis also doesn't seem to need this.

It's not particularly important, but you might want to consider looking at the cause of this in your environment as it could be indicating some setup problem.

nathanegillett: `LambdaClient().handler` isn't the issue
> exodus_lambda/functions/origin_response/origin_response.py:69:0: C0103: Constant name "lambda_handler" doesn't conform to UPPER_CASE naming style (invalid-name)

Not sure what specifically about my environment could be causing the complaint, as it's just the dev-env Python3 virtualenv.

PR #38, opened 2020-04-02
Implement template for bucket and table resources [DELIVERY-8987]

nathanegillett:
This commit introduces a template used to configure and deploy all
necessary resources for an S3 bucket and DynamoDB table.

### DIFF
--- a/configuration/exodus-storage.yaml
+++ b/configuration/exodus-storage.yaml
@@ -1,0 +1,91 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Description: Configuration for exodus-cdn storage resources
+
+Parameters:
+  env:
+    Type: String
+    AllowedValues:
+      - dev
+      - stage
+      - prod
+    Default: dev
+    Description: The environment for which these resources are intended
+
+Resources:
+  Table:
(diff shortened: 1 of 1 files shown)


nathanegillett:
This template was used to create the following stack;
https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/stackinfo?stackId=arn%3Aaws%3Acloudformation%3Aus-east-1%3A333296168829%3Astack%2Fexodus-cdn-dev%2F8e36fd80-750b-11ea-8ffc-0a3238f55eb3

configuration/exodus-storage.yaml
```diff
@@ -0,0 +1,91 @@
...
+      KeySchema:
+        -
+          AttributeName: web_uri
+          KeyType: HASH
+        -
+          AttributeName: from_date
+          KeyType: RANGE
+      BillingMode: PROVISIONED
+      ProvisionedThroughput:
+        ReadCapacityUnits: 10
```
rohanpm, line 37: I reckon these may end up parametrized as well.  Dev environment should probably have far less capacity units than prod, for example, to avoid unnecessary costs.

PR #39, opened 2020-04-03
Resolve RHUI path aliases, via cdn-definitions

rohanpm:
Requests for repos under a /rhui/ path alias should be routed to the
same repo under non-/rhui/ path (e.g. /content/dist/rhel/rhui/foobar
is the same as /content/dist/rhel/foobar).

There is one exception: listing files should not be subject to this
aliasing, because different $releasever or $basearch can be
available for RHUI vs non-RHUI.

Includes changes to alias resolution mechanism:

- aliases now come from the cdn-definitions project, a newly
  introduced dependency, rather than being maintained in the
  config file of this project.  Dependabot is configured to
  submit PRs for all new releases to that project.

- alias resolution was rewritten to no longer depend on the order
  in which aliases are provided.

A couple of extra tests were added for some edge cases.


requirements.txt
```diff
@@ -6,3 +6,4 @@ python-dateutil==2.8.1 --hash=sha256:75bb3f31ea686f1197762692a9ee6a7550b59fc6ca3
 s3transfer==0.3.3 --hash=sha256:2482b4259524933a022d59da830f51bd746db62f047d6eb213f2f8855dcb8a13
 six==1.14.0 --hash=sha256:8f3cd2e254d8f793e7f3d6d9df77b92252b52637291d0f0da013c76ea2724b6c
 urllib3==1.25.8 --hash=sha256:2f3db8b19923a873b3e5256dc9c2dedfa883e33d87c690d9c7913e1f40673cdc
+cdn-definitions==0.2.0 --hash=sha256:2e4ec8b8c9fb743d1295ce979a855b726adf06cf67a27d8217f190ce1ae4c718
```
rohanpm, line 9: I'm planning to do a new release of cdn-definitions project soon after this is landed to validate that dependabot makes an appropriate PR.

rohanpm:
Updated to resolve conflicts (actually, was auto-resolved by git).

PR #40, opened 2020-04-03
Introduce GPG whitelist [RHELDST-14]

nathanegillett:
The deployment pipeline needs to import public keys of trusted entities
to verify signed tags pushed to the "deploy" branch. This commit
introduces a whitelist from which the trusted public keys will be
sourced.


nathanegillett:
Updated commit message with new Jira key

PR #41, opened 2020-04-07
autotests for [DELIVERY-8986]

dichen16:
autotests for DELIVERY-8986: "Exodus CDN should serve repo entry points with appropriate cache headers"

### DIFF
--- a/tests/integration/config.yml
+++ b/tests/integration/config.yml
@@ -1,0 +1,11 @@
+# config.yml for integration test of exodus lambda function
+
+dynamodb_table : 'dichen-table'
+
+cloudfront: 'https://d2dz0xukgxp0w5.cloudfront.net'
+
+patterns:
+  - '/PULP_MANIFEST'
+  - '/listing'
+  - '/repodata/repomd.xml'
+  - '/ostree/repo/refs/heads/'
--- a/tests/integration/test_exoduslambda.py
+++ b/tests/integration/test_exoduslambda.py
@@ -1,0 +1,33 @@
+import pytest
+import logging
+import requests
+
+from .utils import get_config, get_db_client
+from boto3.dynamodb.conditions import Key, Attr
+
+
+class TestExodus:
+    def setup(self):
+        self.config = get_config()
+        self.db_client = get_db_client()
+
+    def test_cache_headers(self):
+        table = self.db_client.Table(self.config["dynamodb_table"])
+        for pattern in self.config["patterns"]:
+            # TODO: update "contains" ComparisonOperator to match real condition
+            response = table.scan(
+                FilterExpression=Attr("web_uri").contains(pattern)
+            )
+
+            # Concatenate URI matching the pattern
+            uri = self.config["cloudfront"] + response["Items"][0]["web_uri"]
+            r = requests.get(uri)
+
+            assert r.status_code == 200
+            assert r.headers["cache-control"] == "max-age=600"
+
+    def test_no_cache_headers(self):
+        r = requests.get(self.config["cloudfront"] + "/no_exist_file")
+
+        assert r.status_code == 404
+        assert "cache-control" not in r.headers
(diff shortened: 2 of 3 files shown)


tests/integration/utils.py
```diff
@@ -0,0 +1,12 @@
+import yaml
```
rohanpm, line 1: You'll need to add PyYAML to test-requirements.txt to fix the current CI failure.

tests/integration/config.yml
```diff
@@ -0,0 +1,11 @@
+# config.yml for integration test of exodus lambda function
```
rohanpm, line 1: I don't think this should require a config file checked into the repo... or at least, it should not be the only way to configure the tests.  It's important to be able to run the same test suite against different environments (e.g. production, a local dev env, or staging env after a deployment was put there).

Also, these tests shouldn't be run by default in CI or when someone runs "pytest" or "tox" locally.  They probably are, at the moment.

I think these two things are related.  Here's a suggestion to fix both issues:

1. Use pytest_addoption feature to add a new option to pytest command which accepts the base URL of the CDN.

2. Set up integration tests so that they skip unless that option is provided.

Overall that would look something like this - in conftest.py (not tested):

```python
def pytest_addoption(parser):
    parser.addoption("--cdn-test-url", action="store", help="enable integration tests against this CDN")

@pytest.fixture
def cdn_test_url(request):
    out = request.config.getoption("--cdn-test-url")
    if not out:
      pytest.skip("Test skipped without --cdn-test-url")
    return out
```

Then each integration test, if it accepts cdn_test_url as an argument, will be skipped if --cdn-test-url is not provided.

tests/integration/test_exoduslambda.py
```diff
@@ -0,0 +1,33 @@
...
+            )
+
+            # Concatenate URI matching the pattern
+            uri = self.config["cloudfront"] + response["Items"][0]["web_uri"]
+            r = requests.get(uri)
+
+            assert r.status_code == 200
+            assert r.headers["cache-control"] == "max-age=600"
+
+    def test_no_cache_headers(self):
```
rohanpm, line 29: This tests that cache headers are missing for a 404 response, there should be a similar test for a 200 response.

tests/integration/test_exoduslambda.py
```diff
@@ -0,0 +1,33 @@
...
+import requests
+
+from .utils import get_config, get_db_client
+from boto3.dynamodb.conditions import Key, Attr
+
+
+class TestExodus:
+    def setup(self):
+        self.config = get_config()
+        self.db_client = get_db_client()
```
rohanpm, line 12: Can we please rewrite this in a way which doesn't require access to the DB?  I would like these tests to be usable from the outside without any special permissions, as long as you have the ability to fetch from the CDN.

Using the DB to find URLs for test in this way is also kinda risky since you could potentially get a different URL each time, leading to unstable test results.  And expensive, at least the way you're doing it now (table scan means reading every row in the table, and this table will be huge on production).

My suggestion is, instead of using the DB to find URLs for test, please simply hardcode the paths for test.  We can call those paths the "reference testdata" and require that every environment has that data in order to support these tests.

dichen16: I believe the "reference testdata" can make same test suite capable for different environments and help manage the test_data, but it also limit the test coverage. I am not sure about it. Is using DB really that unacceptable?

rohanpm: Yeah, you really shouldn't use the DB to drive these tests.

I don't see how it limits the test coverage to avoid using the DB - you don't even know which URLs are tested with the current approach, do you?  So you don't actually know what's covered right now...

As an example of the flaws with the DB-driven approach: the lambda only responds with data having a from_date in the past.  Your test code querying the DB didn't take this into account, so the test could find URLs which are not yet exposed and then the test would incorrectly fail.  To "fix" that with current approach, you'd essentially need the test to reproduce the same logic as used within the lambda and then maintain the two of them together.  I don't think that makes sense... it is too risky that the same bug is written into both the test and the code under test.

PR #43, opened 2020-04-07
Implement template for Lambda@Edge deployment [RHELDST-8]

nathanegillett:
This commit introduces a template used by AWS CloudFormation to
create/update the resources for Lamda@Edge applications.

Those resources include; AWS CloudFront Distribution, AWS Lambda@Edge
functions, AWS IAM Roles for each function, and AWS CloudWatch Alarms
for each function.

The template takes a parameter for the environment -- dev, stage, or
prod -- and a parameter for the environment's origin access identity ID,
for which the exodus-storage template is responsible for creating.

### DIFF
--- a/configuration/exodus-lambda-deploy.yaml
+++ b/configuration/exodus-lambda-deploy.yaml
@@ -1,0 +1,148 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Transform: AWS::Serverless-2016-10-31
+
+Description: Configuration for exodus-cdn Lambda@Edge deployment
+
+Parameters:
+  env:
+    Type: String
+    AllowedValues:
+      - dev
+      - stage
+      - prod
+    Default: dev
+    Description: The environment in which to deploy functions
+  oai:
+    Type: String
+    MinLength: 10
+    Description: The origin access identity ID associated with the environment
+
+Resources:
+  Distribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        Comment: !Sub exodus-cdn-${env}
+        DefaultCacheBehavior:
+          AllowedMethods:
+            - GET
+            - HEAD
+          CachedMethods:
+            - GET
+            - HEAD
+          ForwardedValues:
+            QueryString: false
+            Cookies:
+              Forward: none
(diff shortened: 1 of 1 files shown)


configuration/exodus-lambda-deploy.yaml
```diff
@@ -0,0 +1,148 @@
...
+      Namespace: AWS/Lambda
+      MetricName: Errors
+      Dimensions:
+        - Name: FunctionName
+          Value: !Ref OriginRequestFunc
+      Statistic: Sum
+      Period: 180
+      EvaluationPeriods: 1
+      ComparisonOperator: GreaterThanOrEqualToThreshold
+      Threshold: 3
```
rohanpm, line 111: I have little idea if this is correct.  I won't be surprised if, at the scale we're expecting, we get many many more errors than this all the time - given that clients can probably trigger errors by bad behavior on their end.  Anyway I don't see much choice but to start with a guess and refine it over time.

configuration/exodus-lambda-deploy.yaml
```diff
@@ -0,0 +1,148 @@
...
+        - arn:aws:iam::aws:policy/AWSLambdaInvocation-DynamoDB
+      Path: /
+
+  OriginRequestAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmName: !Sub origin-request-alarm-${env}
+      AlarmDescription: !Sub origin-request-${env} invocation errors
+      Namespace: AWS/Lambda
+      MetricName: Errors
```
rohanpm, line 103: Do you know what kind of "errors" this means?  Is it only when the lambda fails to return a value (e.g. raises uncaught exception) or does it also include HTTP-level errors like a 404, which can happen even when lambda is working perfectly?

nathanegillett: I believe it's only concerned with errors raised by the code

configuration/exodus-lambda-deploy.yaml
nathanegillett: You're probably right. The idea behind this threshold was for it to cause a rollback of the CloudFormation stack during validation. So, a few failed pipeline tests should be enough to keep out bad code.

PR #45, opened 2020-04-14
Add buildspec.yml for AWS CodeBuild [RHELDST-20, RHELDST-24]

crungehottman:
This commit introduces a buildspec.yml file to provide to AWS CodeBuild during
the build phase of the exodus CodePipeline. CodeBuild packages the exodus Lambda@Edge
code from the latest verified git tag into a deployable CloudFormation template.

Additionally, the Lambda@Edge function configuration is dynamically generated during the Build phase.

### DIFF
--- a/configuration/buildspec.yaml
+++ b/configuration/buildspec.yaml
@@ -1,0 +1,11 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - ./scripts/build-package
+
+artifacts:
+  files:
+    - ./package/exodus-lambda-pkg.yaml
+  discard-paths: yes
--- a/scripts/build-package
+++ b/scripts/build-package
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -e
+
+pip install --require-hashes -r requirements.txt --target ./package
+pip install --no-deps --target ./package .
+cp ./configuration/exodus-lambda-deploy.yaml ./package
+envsubst < ./configuration/lambda_config.json > ./package/lambda_config.json
+aws cloudformation package \
+	--template ./package/exodus-lambda-deploy.yaml \
+	--s3-bucket exodus-pipeline-artifacts \
+	--output-template-file ./package/exodus-lambda-pkg.yaml
--- a/configuration/lambda_config.json
+++ b/configuration/lambda_config.json
@@ -1,7 +1,7 @@
 {
   "table": {
-    "name": "exodus",
-    "region": "us-east-1"
+    "name": "exodus-cdn-$ENV_TYPE",
+    "region": "$AWS_REGION"
   },
   "headers": {
     "max_age": "600"


.whitelist.gpg
```diff
@@ -56,4 +56,56 @@ xRYB7YaXneBhWc7F6cC0mz4mwrcVH952VR9qAKVXK5QR16EZaDumxjB2uH619w68
 oekxZ1I7S/t/zqeQ04vucdhRVYyzyfusjtmwSOLkZDdLfeFxqkB7vnns+yGqkLfq
 WkotIE5Gq09SL6AXJAVvq9osefk4kDYc5zXmFuA1CHpQJMBziWfOC1c7RndBzw==
 =KI3J
------END PGP PUBLIC KEY BLOCK-----
\ No newline at end of file
+-----END PGP PUBLIC KEY BLOCK-----
```
crungehottman, line 59: I will remove my gpg key before the final review and merge, but it is necessary for testing.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - git clone https://github.com/crungehottman/exodus-lambda.git
```
crungehottman, line 6: Before merge/final review, I will change the repo to https://github.com/release-engineering/exodus-lambda, but my fork and test branch are required for testing.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+      - cd exodus-lambda
+      - git checkout codebuild
```
crungehottman, line 8: Before merge/final review, I will remove this line.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+         }" > ./package/lambda_config.json
+      - aws cloudformation package
+        --template ./package/exodus-lambda-deploy.yaml
+        --s3-bucket exodus-pipeline-artifacts
+        --output-template-file ./package/exodus-lambda-pkg.yaml
+
+artifacts:
+  files:
+    - ./exodus-lambda/package/lambda_config.json
+    - ./exodus-lambda/package/exodus-lambda-pkg.yaml
```
nathanegillett, line 33: We shouldn't need lambda_config.json exported. It's already in the package and available to the function.
Also, it would be nice to have `discard-paths` option here so we don't need to use the whole path when referencing files later on.
```suggestion
artifacts:
  files:
    - ./exodus-lambda/package/exodus-lambda-pkg.yaml
  discard-paths: yes
```

.whitelist.gpg
nathanegillett: I don't have a problem with your key being on the whitelist, myself.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+      - git verify-tag $(git describe $LAST_TAGGED)
+      - git checkout $LAST_TAGGED
+      - pip install --require-hashes -r requirements.txt --target ./package
+      - pip install --no-deps --target ./package .
+      - cp ./configuration/exodus-lambda-deploy.yaml ./package
+      - >-
+        jq -n "{
+           \"table\": {
+             \"name\": \"exodus-cdn-$ENV_TYPE\", region: \"us-east-1\"
+           },
```
nathanegillett, line 20: This wasn't mentioned anywhere but I think region should be an environment variable also. The pipeline will need to be aware of the AWS region anyway so allowing it to fill this value like it will elsewhere can protect against inconsistencies.

crungehottman:
> Does the exodus-lambda-pkg.yaml generated work to deploy usable Lamba@Edge functions?
> 
> Building the package with AWS cli command only seemed to work for me if I did so from within the package directory. Otherwise I had strange nesting/pathing issues, IIRC.

I was able to build and deploy some Lambda@Edge functions using the following commands: 
`aws cloudformation package --template ./package/exodus-storage.yaml --s3-bucket exodus-pipeline-artifacts --output-template-file ./package/exodus-storage-pkg.yaml`
and
`aws cloudformation deploy --template-file /home/crungeho/git/exodus-lambda/package/exodus-lambda-pkg.yaml --stack caleigh-test-1 --parameter-overrides oai=$MYOAI env=stage --capabilities CAPABILITY_NAMED_IAM`

However, I have not tried adding content to my bucket to verify the lambdas' usability. That will be my next test.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+      - git checkout codebuild
+      - gpg --import .whitelist.gpg
+      - LAST_TAGGED=$(git rev-list --tags --max-count=1)
+      - git verify-tag $(git describe $LAST_TAGGED)
+      - git checkout $LAST_TAGGED
+      - pip install --require-hashes -r requirements.txt --target ./package
+      - pip install --no-deps --target ./package .
+      - cp ./configuration/exodus-lambda-deploy.yaml ./package
+      - >-
+        jq -n "{
```
rohanpm, line 17: Could we find a way to have this template in a separate file in the repo rather than embedded into this YAML?

It would be neat if the template file could use YAML & Jinja for consistency with ansible.

rohanpm: Actually, taking it further: I think everything here after the git verification and checkout steps should not be embedded directly in the buildspec.yml.  They could instead be hosted in a script in the repo (so that the buildspec would just do: verify & checkout desired revision, then run e.g. "scripts/build-package" or whatever).

The reason is: we didn't design anything to automatically deploy this buildspec.yml into AWS, did we?  Does that mean every time you need to change it, someone is expected to manually run some commands locally to deploy it?

If so, we should try to have the narrowest possible scope in the buildspec so the need for such changes would be minimized, because every change needs some special action beyond the usual release workflow, which is prone to human error (e.g. a developer accidentally deploying some locally modified version of the buildspec.yml rather than what's in git).

So, for example, if we need to change some detail of the config file template, or slightly change the "aws cloudformation ..." command which is run, ideally we can achieve that by the same workflow as used for changes to the lambdas themselves: merge to deployment branch and make a new signed tag.  No "deploy updated buildspec" step which could be forgotten or executed incorrectly.

nathanegillett: > Does that mean every time you need to change it, someone is expected to manually run some commands locally to deploy it?
No, changes to the buildspec would be applied when pushed to the deploy branch, like any other change to exodus-lambda source.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - git clone https://github.com/crungehottman/exodus-lambda.git
+      - cd exodus-lambda
+      - git checkout codebuild
+      - gpg --import .whitelist.gpg
```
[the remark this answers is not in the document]
nathanegillett: We could use the whitelist from a previous tag. Though that doesn't offer much more protection. Someone could just push two tagged commits, the first containing their key.

Originally, I thought of importing keys from distribution-password-store, but that would require cloning yet another repo and traversing the pubkey files. Anyway, distribution-password-store could potentially be hijacked too, right?

How many layers are enough?

Would importing from a key server be better somehow? We'd still need to include trusted email addresses somewhere for that, so it would seem to carry the same flaws as the current approach.

buildspec.yml
```diff
@@ -0,0 +1,33 @@
...
+      - LAST_TAGGED=$(git rev-list --tags --max-count=1)
```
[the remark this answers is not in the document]
nathanegillett: It's starting to sound to me like we may prefer a manual deployment process over the automated pipelines.

We would have to checkout whatever source we want to deploy and use CloudFormation cli to package and deploy the Lambda@Edge template. Alarms could still be used to automatically rollback in the event of _n_ function failures but automated testing would have to be setup differently, perhaps detecting the change in live function versions somehow.

buildspec.yml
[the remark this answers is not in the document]
crungehottman: > We could use the whitelist from a previous tag. Though that doesn't offer much more protection. Someone could just push two tagged commits, the first containing their key.

+1

Another potential approach: Instead of pulling the whitelist directly from git, the whitelist could exist as an artifact in our exodus-pipeline-artifacts bucket. CodeBuild would use the whitelist as a source artifact provided to the build to be used to import and verify the GPG keys. The whitelist from the git tag would exit the build phase as a build artifact. If the key used to sign the tag is in the whitelist, the build passes and the pipeline continues into a new 'replace whitelist' phase, which replaces the whitelist in the bucket with the whitelist from the git tag. If the GPG key used to sign the git tag is not in the whitelist, the build fails and the rest of the pipeline will not execute, including the replacement of the gpg whitelist -- the list from the previous successful deployment is maintained.

scripts/build-package
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+pip install --require-hashes -r requirements.txt --target ./package
+pip install --no-deps --target ./package .
+cp ./configuration/exodus-lambda-deploy.yaml ./package
+envsubst < ./configuration/lambda_config.json > ./package/lambda_config.json
+aws cloudformation package \
+	--template ./package/exodus-lambda-deploy.yaml \
+       	--s3-bucket exodus-pipeline-artifacts \
+	--output-template-file ./package/exodus-lambda-pkg.yaml
```
nathanegillett, line 11:
```suggestion
aws cloudformation package \
	--template ./package/exodus-lambda-deploy.yaml \
	--s3-bucket exodus-pipeline-artifacts \
	--output-template-file ./package/exodus-lambda-pkg.yaml
```
Ignore if this is only present in the Github view

buildspec.yml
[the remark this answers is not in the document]
rohanpm: bq. We would have to checkout whatever source we want to deploy and use CloudFormation cli to package and deploy the Lambda@Edge template.

You mean check it out locally?  I don't think it's a good idea.  Every time that is done, there's a risk that the person doing it is accidentally deploying something wrong, like untested local changes, some hacks they added for development and forgot to remove, etc.... it's not sufficiently under control.

buildspec.yml
rohanpm: So, just to be clear then: the repo is checked out more than once?  At least once to deploy the buildspec and then again while running the buildspec?

If buildspec changes checked into the repo are applied, then it's futile to try to add any kind of verification into the buildspec steps itself as the person who managed to inject their malicious commit can just strip out those verification steps from the buildspec.

buildspec.yml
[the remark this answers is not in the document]
rohanpm: > Originally, I thought of importing keys from distribution-password-store, but that would require cloning yet another repo and traversing the pubkey files. Anyway, distribution-password-store could potentially be hijacked too, right?

I don't consider that repo being hijacked as in scope as it is Red Hat internal.  However I don't think you can use it for that exact reason.  Also, based on the other conversation thread about how the buildspec changes get deployed, if you add such checks in the buildspec, they can simply be removed by the attacker, right?  Same goes for *any* verification added in the buildspec for that matter.

The threat model I want us to be robust against has assumptions:

1. Assume that *this repo* can be written to by an attacker
2. Assume anything already deployed in AWS is not compromised (if it was then the game is already lost)
3. Assume Red Hat internal repos & access control mechanisms are not compromised (though I'm not sure it's relevant)

Now imagine we have a CodePipeline deployed in AWS.  I read Nathan's comments in another thread as meaning: when that pipeline is used to deploy revision X, the build runs whatever build steps are in the buildspec file as of revision X.  That makes it futile to use the buildspec to verify what's being deployed, because the buildspec itself can be written to by an attacker.

If this understanding of the buildspec is correct then it seems like the design from RHELDST-20 asking to have GPG verification in the buildspec is wrong, it can't be done usefully here.  It could possibly be done usefully earlier in the pipeline though, but that also depends on how the pipeline is deployed.  This would be a lot easier to judge if we already had the pipeline to review as well.

Does this make sense?  If so I would recommend stripping out the GPG verification stuff here and instead shifting that  into the scope of RHELDST-25.

buildspec.yml
nathanegillett: > [...] the repo is checked out more than once?

No, the pipeline receives a zip of the repo's files, in which the buildspec is included, whenever a commit is pushed to the deploy branch. It's this copy of the buildspec that's used to clone and checkout the full repo.

> [...] it's futile to try to add any kind of verification into the buildspec steps itself as the person who managed to inject their malicious commit can just strip out those verification steps from the buildspec.

Yeah...this was an oversight. I'm for "stripping out the GPG verification stuff here and instead shifting that into the scope of RHELDST-25."

buildspec.yml
```diff
@@ -0,0 +1,18 @@
+version: 0.2
```
rohanpm, line 1: About the name & location of this file:

- it's named *.yml but other AWS config files are named *.yaml
- it's placed in the root dir but other AWS config files are placed in configuration dir

If there's no reason for the inconsistency, can you please make it consistent?

buildspec.yml
rohanpm: Is it meant to be removed, or changed to "deploy" branch?  That's what https://github.com/release-engineering/exodus-lambda/pull/48/files refers to.

configuration/buildspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+phases:
+  install:
+    commands:
+      - yum install gettext -y
+
+  build:
+    commands:
+      - git clone https://github.com/release-engineering/exodus-lambda.git
+      - cd exodus-lambda
+      - git checkout deploy
```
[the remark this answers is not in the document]
crungehottman: I agree. Cloning the repo at build-time was only necessary for verifying the GPG key signature. I will use the pipeline source instead.

configuration/buildspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+phases:
+  install:
+    commands:
+      - yum install gettext -y
+
+  build:
+    commands:
+      #- git clone https://github.com/release-engineering/exodus-lambda.git
+      - cd $CODEBUILD_SRC_DIR
+      #- git checkout deploy
```
nathanegillett, line 12: I think all of this can be removed. Build should already be at root of the repo, exodus-lambda/.

configuration/buildspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - ./scripts/build-package
+
+artifacts:
+  files:
+    - ./exodus-lambda/package/exodus-lambda-pkg.yaml
```
nathanegillett, line 17: Curious if this could just be `./package/exodus-lambda-pkg.yaml`.

.whitelist.gpg
```diff
@@ -56,4 +56,5 @@ xRYB7YaXneBhWc7F6cC0mz4mwrcVH952VR9qAKVXK5QR16EZaDumxjB2uH619w68
...
+
```
nathanegillett, line 60: Extra newline?

scripts/build-package
```diff
@@ -0,0 +1,13 @@
...
+pip install --require-hashes -r requirements.txt --target ./tmp_package
+pip install --no-deps --target ./tmp_package .
+cp ./configuration/exodus-lambda-deploy.yaml ./tmp_package
+envsubst < ./configuration/lambda_config.json > ./tmp_package/lambda_config.json
+cd ./tmp_package
+ls
+aws cloudformation package \
+	--template exodus-lambda-deploy.yaml \
+	--s3-bucket codepipeline-us-east-1-956509269028 \
+	--output-template-file exodus-lambda-pkg.yaml
```
nathanegillett, line 13: Just a reminder, s3-bucket should point to "exodus-pipeline-artifacts" before merge

PR #46, opened 2020-04-16
integration test [RHELDST-9]

dichen16:
The test data in this commit is only work for specific `-cdn-test-url`,
`--cdn-test-url https://d2dz0xukgxp0w5.cloudfront.net`, 
test data will be optimized in RHELDST-590.

automate integration test for RHELDST-13, 22, 23.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,7 +20,7 @@ whitelist_externals=
     cp
 commands_pre=
 commands=
-    rm -rf ./package
+    rm -rf ./package package.zip
 
     # --require-hashes and then --no-deps to avoid using anything untrusted from PyPI
     pip install --require-hashes -r requirements.txt --target ./package


dichen16:
there is a useless commit in this patch, once the patch meet the merge requirement, I will re-push a clean patch.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,81 @@
...
+    assert r.status_code == 200
+    assert "digest" in r.headers
+
+
+def test_header_want_digest_head(cdn_test_url):
+    headers = {"want-digest": "sha256"}
+    url = cdn_test_url + "/content/beta/rhel8/8/x86_64/exodus_test/os/listing"
+    r = requests.head(url, headers=headers)
+    assert r.status_code == 200
+    assert "digest" in r.headers
```
nathanegillett, line 51: I think we should verify the digests in these tests since our code is responsible for creating the header.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_header_not_exist_file(cdn_test_url):
+    url = (
+        cdn_test_url
+        + "/content/beta/rhel8/8/x86_64/exodus_test/os/bash-4.4.19-7.el8.x86_64.rpm_no"
+    )
+    r = requests.get(url)
+    assert r.status_code == 404
+    assert "cache-control" not in r.headers
```
nathanegillett, line 22: Harmless, I suppose, but I don't believe it's possible for custom response headers to be applied on 404.

rohanpm: I think it is - our lambdas apparently can customize error responses, https://aws.amazon.com/about-aws/whats-new/2017/12/lambda-at-edge-now-allows-you-to-customize-error-responses-from-your-origin/, so it does seem possible that we could accidentally leave headers in for error cases.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_header_cache_control(cdn_test_url):
+    patterns = [
```
rohanpm, line 26: Can you please use pytest.mark.parametrize instead of embedding the data in the test like this?

One reason is that you'll get improved failure messages that way.  For instance, right now if this test fails on the r.status_code == 200 check, it won't tell you *which* URL it failed on.  If you use test parameters instead, it will tell you.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,81 @@
...
+    ]
+    multiple_url = [cdn_test_url + pattern_item for pattern_item in patterns]
+    for url in multiple_url:
+        r = requests.get(url)
+        assert r.status_code == 200
+        assert r.headers["cache-control"] == "max-age=600"
+
+
+def test_header_want_digest_get(cdn_test_url):
+    headers = {"want-digest": "sha256"}
```
rohanpm, line 39: This isn't the want-digest we expect users to pass, they should pass id-sha-256.  It doesn't actually make a difference to the test because we still return id-sha-256 anyway.  I'd prefer if you could provide the real checksum type though.

Actually I'm getting a bit nervous about our behavior of returning id-sha-256 even if the user passed something wrong; I might file an issue requesting to change that.

tests/integration/test_exodus.py
rohanpm: I agree.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,81 @@
...
+    assert r.status_code == 200
+    assert "digest" in r.headers
+
+
+def test_repodata_content_type_xml(cdn_test_url):
+    url = (
+        cdn_test_url
+        + "/origin/files/sha256/5c/5c50cbb468bf0bde6587019e5ba51346506793a65e914d06c6f98b4c9c0a1598/repomd.xml"
+    )
+    r = requests.get(url)
```
rohanpm, line 59: These tests should really be done with both GET and HEAD since the content-type is supposed to be the same for both.  You could parametrize that.

tests/integration/test_exoduslambda.py
```diff
@@ -0,0 +1,33 @@
...
+from .utils import get_config, get_db_client
+from boto3.dynamodb.conditions import Key, Attr
+
+
+class TestExodus:
+    def setup(self):
+        self.config = get_config()
+        self.db_client = get_db_client()
+
+    def test_cache_headers(self):
```
rohanpm, line 14: Could we please not have this test added, and instead stick with the other one only?

dichen16: yes

tests/integration/test_exodus.py
dichen16: @rohanpm ,  how about this?
```
testdata_content_type_path = [
    "/content/beta/rhel8/8/x86_64/exodus_test_ctype_2/os/repodata/acc399cea3af512ef5f773604f0ca7c129c6a973c6364786954ec49e88f3582f-filelists.sqlite.bz2",
    "/content/beta/rhel8/8/x86_64/exodus_test_ctype_2/os/repodata/47e27d9fff2396c32b4a6c1462282fdb5c8330a14c8ae647e39242cf5c5281ce-primary.xml.gz",
    "/origin/files/sha256/5c/5c50cbb468bf0bde6587019e5ba51346506793a65e914d06c6f98b4c9c0a1598/repomd.xml"
]

def assert_content_type(url, content_type):
    if url.endswith("repomd.xml"):
        assert content_type == "application/xml"
    elif url.endswith(".gz"):
        assert content_type == "application/x-gzip"
    elif url.endswith(".bz2"):
        assert content_type == "application/x-bzip"
    else:
        raise ValueError("invalid test data format")

@pytest.mark.parametrize("testdata_path", testdata_content_type_path)
def test_content_type_header_GET(cdn_test_url, testdata_path):
    url = cdn_test_url + testdata_path
    r = requests.get(url)
    assert r.status_code == 200
    assert_content_type(url, r.headers["content-type"])

@pytest.mark.parametrize("testdata_path", testdata_content_type_path)
def test_content_type_header_HEAD(cdn_test_url, testdata_path):
    url = cdn_test_url + testdata_path
    r = requests.head(url)
    assert r.status_code == 200
    assert_content_type(url, r.headers["content-type"])
```

PR #47, opened 2020-04-17
integration test for RHELDST-13, 22, 23

dichen16:
The test data in this commit is only work for specific -cdn-test-url,
--cdn-test-url https://d2dz0xukgxp0w5.cloudfront.net,
Test data will be optimized in RHELDST-590.

automate integration test for RHELDST-13, 22, 23.

### DIFF
--- a/tests/integration/test_exodus.py
+++ b/tests/integration/test_exodus.py
@@ -1,0 +1,98 @@
+import requests
+import pytest
+
+
+def test_exodus_basic(cdn_test_url):
+    url = (
+        cdn_test_url
+        + "/content/beta/rhel8/8/x86_64/exodus_test/os/bash-4.4.19-7.el8.x86_64.rpm"
+    )
+
+    r = requests.get(url)
+    assert r.status_code == 200
+    assert "cache-control" not in r.headers
+
+
+def test_header_not_exist_file(cdn_test_url):
+    url = (
+        cdn_test_url
+        + "/content/beta/rhel8/8/x86_64/exodus_test/os/bash-4.4.19-7.el8.x86_64.rpm_no"
+    )
+    r = requests.get(url)
+    assert r.status_code == 404
+    assert "cache-control" not in r.headers
+
+
+# Test data for exodus-lambda
+# serving repo entry points with appropriate cache headers
+testdata_cache_control_path = [
+    "/content/beta/rhel8/8/x86_64/exodus_test/os/listing",
+    "/content/beta/rhel8/8/x86_64/exodus_test_ctype/os/repodata/repomd.xml",
+    "/content/beta/rhel8/8/x86_64/exodus_test/os/ostree/repo/refs/heads/sub_dir/testing",
+]
+
+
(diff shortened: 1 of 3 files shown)


tests/integration/test_exodus.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_header_want_digest_HEAD(cdn_test_url):
+    headers = {"want-digest": "id-sha-256"}
+    url = cdn_test_url + "/content/beta/rhel8/8/x86_64/exodus_test/os/listing"
+    r = requests.head(url, headers=headers)
+    assert r.status_code == 200
+    assert (
+        r.headers["digest"]
+        == "id-sha-256=jbuWD7I1f2C3IcJT6acQqIz1kI/pffRTD3krSieXQQc="
+    )
```
nathanegillett, line 62: I'd prefer we check the decoded the digests

tests/integration/conftest.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+
+def pytest_addoption(parser):
```
rohanpm, line 4: In order for this to work in CI, you'll apparently need to put the file at tests/conftest.py instead.  I think what happens is that pytest finds `testpaths = tests` in tox.ini and then only processes tests/conftest.py prior to handling command-line options.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,98 @@
+import requests
```
rohanpm, line 1: Please add requests into test-requirements.txt as well, as nothing guarantees that it's installed.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,98 @@
...
+    url = (
+        cdn_test_url
+        + "/content/beta/rhel8/8/x86_64/exodus_test/os/bash-4.4.19-7.el8.x86_64.rpm_no"
+    )
+    r = requests.get(url)
+    assert r.status_code == 404
+    assert "cache-control" not in r.headers
+
+
+# Test data for exdous-lambda
```
rohanpm, line 26:
```suggestion
# Test data for exodus-lambda
```

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,98 @@
...
+# serving yum repodata with correct Content-Type
```
rohanpm, line 27: What's this comment talking about?  It doesn't seem to have any direct connection with the following variable, which is about cache-control.

tests/integration/test_exodus.py
```diff
@@ -0,0 +1,98 @@
...
+    r = requests.head(url, headers=headers)
+    assert r.status_code == 200
+    assert (
+        r.headers["digest"]
+        == "id-sha-256=jbuWD7I1f2C3IcJT6acQqIz1kI/pffRTD3krSieXQQc="
+    )
+
+
+# Test data for exodus-lambda
+# serving repo entry points with appropriate cache headers
```
rohanpm, line 66: Similarly this comment refers to something different from what's tested...

Did two comments get swapped around?

PR #48, opened 2020-04-17
Implement template for CodePipeline [RHELDST-25]

nathanegillett:
This commit introduces a template used by AWS CloudFormation to
create/update the resources for CodePipeline pipelines.

Those resources include; an Amazon SNS topic, CodePipeline pipeline, S3
bucket, CodeBuild project, and the required AWS IAM roles.

The template takes a parameter for the environment -- dev, stage, or
prod -- a parameter for the environment's origin access identity ID, a
parameter for the AWS region, and a parameter for the email address for
notifications.

### DIFF
--- a/configuration/exodus-pipeline.yaml
+++ b/configuration/exodus-pipeline.yaml
@@ -1,0 +1,344 @@
+AWSTemplateFormatVersion: 2010-09-09
+
+Description: Configuration for exodus-cdn CodePipeline
+
+Parameters:
+  env:
+    Type: String
+    AllowedValues:
+      - dev
+      - stage
+      - prod
+    Default: dev
+    Description: The environment in which to deploy functions
+  oai:
+    Type: String
+    Description: The origin access identity ID associated with the environment
+  githubToken:
+    Type: String
+    Default: None
+    Description: A GitHub access token token to authorize webhooks
+  region:
+    Type: String
+    Default: us-east-1
+    Description: The region in which resources are established
+  email:
+    Type: String
+    Default: project-exodus@redhat.com
+    Description: The email address to which notifications are sent
+
+Conditions:
+  IsDev:
+    !Equals [!Ref env, dev]
+  NotDev:
+    !Not [!Equals [!Ref env, dev]]
+  NotProd:
+    !Not [!Equals [!Ref env, prod]]
+
+Mappings:
+  SourceMap:
+    dev:
+      objectKey: ""
+    stage:
+      objectKey: build-dev/exodus-lambda-pkg.yaml
+    prod:
+      objectKey: build-stage/exodus-lambda-pkg.yaml
+
+Resources:
+  SNSTopic:
+    Type: AWS::SNS::Topic
+    Properties:
+      TopicName: !Sub exodus-sns-${env}
+      Subscription:
+        - Endpoint: !Ref email
+          Protocol: email
+
+  CodePipelineNotification:
+    Type: AWS::CodeStarNotifications::NotificationRule
+    Properties:
+      Name: !Sub exodus-pipeline-${env}
+      DetailType: FULL
+      EventTypeIds:
+        - codepipeline-pipeline-pipeline-execution-failed
+      Resource:
+        !Sub arn:aws:codepipeline:${region}:${AWS::AccountId}:${CodePipeline}
+      Targets:
+        - TargetType: SNS
+          TargetAddress: !Sub arn:aws:sns:${region}:${AWS::AccountId}:exodus-sns-${env}
+
+  CodePipelineWebhook:
+    Type: AWS::CodePipeline::Webhook
+    Condition: IsDev
+    Properties:
+      Name: !Sub exodus-pipeline-${env}
+      AuthenticationConfiguration:
+        SecretToken: !Ref githubToken
+      Filters:
+        - JsonPath: "$.ref"
+          MatchEquals: refs/heads/deploy
+      Authentication: GITHUB_HMAC
+      TargetPipeline: !Ref CodePipeline
+      TargetAction: Source
+      TargetPipelineVersion: !GetAtt CodePipeline.Version
+      RegisterWithThirdParty: true
+
+  CodePipeline:
+    Type: AWS::CodePipeline::Pipeline
+    Properties:
+      Name: !Sub exodus-pipeline-${env}
+      ArtifactStore:
+        Type: S3
+        Location: !Ref CodePipelineBucket
+      RoleArn: !GetAtt CodePipelineRole.Arn
+      Stages:
+        - !If
+          # Create Github source for dev pipeline
+          - IsDev
+          - Name: Source
+            Actions:
+              - Name: Source
+                ActionTypeId:
+                  Category: Source
+                  Owner: ThirdParty
+                  Provider: GitHub
+                  Version: 1
+                Configuration:
+                  Owner: release-engineering
+                  Repo: exodus-lambda
+                  Branch: pipeline-template
+                  OAuthToken: !Ref githubToken
+                  PollForSourceChanges: false
+                OutputArtifacts:
+                  - Name: SourceArtifact
+          - !Ref AWS::NoValue
+        - !If
+          # Create S3 source for stage, prod pipelines
+          - NotDev
+          - Name: Source
+            Actions:
+              - Name: Source
+                ActionTypeId:
+                  Category: Source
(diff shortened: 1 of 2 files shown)


nathanegillett:
Pending #45

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,297 @@
...
+                  Version: 1
+                Configuration:
+                  Owner: release-engineering
+                  Repo: exodus-lambda
+                  Branch: deploy
+                  OAuthToken: !Ref github-token
+                  PollForSourceChanges: false
+                OutputArtifacts:
+                  - Name: SourceArtifact
+          # Create S3 source for stage, prod pipelines
```
rohanpm, line 86: Is this comment accurate?  The next branch only seems to be entered if env is prod, not stage.

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,297 @@
...
+        - Name: Deploy
+          Actions:
+            - Name: Deploy
+              ActionTypeId:
+                Category: Deploy
+                Owner: AWS
+                Provider: CloudFormation
+                Version: 1
+              Configuration:
+                ActionMode: CREAT_UPDATE
```
crungehottman, line 138: should this be CREATE_UPDATE?

nathanegillett:
> Can you confirm a couple of things about this:
> 
>     1. There's no attempt to do any kind of GPG verification yet.  Any plan to include that here, or file a separate issue for it?

I think we need to investigate the best way to implement GPG verification first. So, a separate issue seems appropriate.

>     2. I think the intent at the moment is to do something like: dev builds from git; stage takes whatever was built in dev; prod takes whatever was built in stage.  Is that right?

That's the current approach, yeah.

> If so, can we please instead let stage also build from git. The reason is that "dev" is where developers expect to be able to put work in progress, edit/upload resources directly, etc. There should be no chance that a developer can edit something in dev and have that work accidentally promoted into prod.

Hmm. That makes sense. Though, we will need separate branches to achieve the effect you're looking for -- one for dev and another for stage, prod.
Right now dev pipeline is set up to pull from deploy branch. I think this is wrong, given your argument.
If dev pipeline is for WIP stuff, we should should probably source from master branch.
Then pushing a commit to the deploy branch actually has meaning -- this code is stable and ready for deployment to production environments.

configuration/exodus-pipeline.yaml
nathanegillett: Thanks!

configuration/exodus-pipeline.yaml
nathanegillett: Good catch. I originally had stage and prod source stages separated and forgot to change the condition when consolidating.

nathanegillett:
> It would be great if you could write something in docs about this.
> [...]
> Even better would be to have some human-readable explanation of the deployment workflow.

Will do.
Note that this template isn't usable until #45 is merged (and this commit rebased) so I haven't tested it yet.

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,297 @@
...
+    Properties:
+      Name: exodus-codebuild
+      SeviceRole: !GetAtt CodeBuildRole.Arn
+      Environment:
+        Type: LINUX_CONTAINER
+        ComputeType: BUILD_GENERAL1_SMALL
+        Image: aws/codebuild/amazonlinux2-x86_64-standard:2.0
+      Source:
+        Type: CODEPIPELINE
+        BuildSpec: buildspec.yml
```
crungehottman, line 213:
```suggestion
        BuildSpec: configuration/buildspec.yaml
```

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,297 @@
...
+                Region: !Ref region
+                Configuration:
+                  EnvironmentVariables:
+                    - Name: ENV_TYPE
+                      Value: !Ref env
+                      Type: PLAINTEXT
+                    - Name: AWS_REGION
+                      Value: !Ref region
+                      Type: PLAINTEXT
+                  ProjectName: exodus-lambda
```
crungehottman, line 123: The ProjectName referenced in the CodePipeline conflicts with the CodeBuild project name defined in line 205:
```suggestion
                  ProjectName: exodus-codebuild
```

nathanegillett: Thanks!

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,297 @@
...
+      VersioningConfiguration:
+        Status: Enabled
+
+  CodeBuildProject:
+    Type: AWS:CodeBuild:Project
+    # Create CodeBuild project for dev pipeline
+    Condition: !Equals [!Ref env, dev]
+    Properties:
+      Name: exodus-codebuild
+      SeviceRole: !GetAtt CodeBuildRole.Arn
```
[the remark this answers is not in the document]
nathanegillett: Yep, thanks

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,297 @@
...
+                Version: 1
+              Configuration:
+                ActionMode: CREATE_UPDATE
+                Capabilities: CAPABILITY_IAM,CAPABILITY_AUTO_EXPAND
+                ParameterOverrides:
+                  !Join
+                    - '{"env": "'
+                    -
+                      - !Ref env
+                      - '"}'
```
crungehottman, line 145: +oai

nathanegillett:
This template is now working; result at https://console.aws.amazon.com/codesuite/codepipeline/pipelines/exodus-pipeline-dev/view?region=us-east-1

configuration/exodus-pipeline.yaml
```diff
@@ -0,0 +1,345 @@
...
+  #    AuthenticationConfiguration:
+  #      SecretToken: 123456789  # TODO
+  #    Filters:
+  #      - JsonPath: "$.ref"
+  #        MatchEquals: refs/heads/{Branch
+  #    Authentication: GITHUB_HMAC
+  #    TargetPipeline: !Ref CodePipeline
+  #    TargetAction: Source
+  #    TargetPipelineVersion: !GetAtt CodePipeline.Version
+  #    RegisterWithThirdParty: true
```
nathanegillett, line 84: @rohanpm I suspect we'll need this webhook setup, which will require GitHub settings I don't have access to.

rohanpm: @nathanegillett I read the AWS docs relating to this.  Can you confirm my understanding of what's going on here:

- This resource causes AWS to create a webhook on Github, to call back to AWS when branches are changed.
- When authenticating with github to create this hook, AWS uses the OAuthToken from the Source action (i.e. "githubToken" parameter)
- That token needs to have "admin:repo_hook" scope (got this from https://docs.aws.amazon.com/codepipeline/latest/userguide/pipelines-webhooks-create-cfn.html).  That doc also implies it needs "repo" scope but I have a feeling it might not actually be correct for public repos.
- I guess you're using a token you created via your own account while working on this?

If that's all accurate: could you try using the token I just added to distribution-password-store named exodus-aws-oauth?  It has admin:repo_hook scope.  I did not add repo scope since, as I mentioned above, I think it might not actually be needed.

nathanegillett: That all sounds right. I've confirmed that only repo_hook is needed using a personal access token I generated and it works to create the webhook on my fork. I'll continue with this for now until the template is finalized; I'm still trying to work the CloudWatch alarms into deploy stage's CloudFormation for automatic rollbacks.

Thanks for adding the token to password-store. We can use that for actual pipeline deployments in the future.

nathanegillett:
I know it's possible to supply the CloudWatch alarm to the build stage's CloudFormation action via ParameterOverrides (for automatic rollback) but I can't seem to find the reference for it. It's been weeks since I saw it last.

Anyway, after adding the alarm manually via the console, it seems that this feature is blocking -- a monitoring time of 1 hour means that the pipeline cannot proceed past the deploy stage until that hour has lapsed. So, as integration tests won't execute during the rollback window, I think the purpose of this is defeated. Perhaps it's best to leave monitoring/rollback to the tests afterall.

nathanegillett:
I've enabled the GitHub webhook and pointed it and the source stage at release-engineering/exodus-lambda/deploy.

PR #50, opened 2020-04-24
Remove unused files

nathanegillett:
This commit removes CHANGELOG.md and requirements-dev.txt files, as they
are entirely unused.


CHANGELOG.md
```diff
@@ -1,16 +0,0 @@
-# Changelog
```
nathanegillett, line 1: A changelog might be useful in the future, when the exodus cdn is live.

CHANGELOG.md
```diff
@@ -1,16 +0,0 @@
...
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
-
-## [Unreleased]
-
-- n/a
-
-## 0.1.0 - 2020-02-06
-
-- Initial release to PyPI
```
nathanegillett, line 14: There are no plans to release to PyPI right now.

CHANGELOG.md
```diff
@@ -1,16 +0,0 @@
...
-
-[Unreleased]: https://github.com/release-engineering/pushcollector/compare/v0.1.0...HEAD
```
nathanegillett, line 16: Wrong link.

rohanpm:
> requirements-dev.txt probably shouldn't be unused, we support pre-commit in this repo and I think requirements-dev.txt would be a typical place to put requirements needed to run the repo's precommit hooks. However, it can be restored later if someone wants to make it non-empty.

Actually our README even refers to requirements-dev.txt in this way, so with this change our README contains instructions for setting up pre-commit which don't work.

PR #51, opened 2020-04-24
Correct pipeline template for stage,prod environments

nathanegillett:
Previously, the template for creating pipelines only worked for the
initial, dev pipeline. This was due to stage and prod pipelines
attempting to reference resourses only created in the dev environment.
This commit replaces those references with the reousrce ARNs.

Additionally, the final pipeline action was changed to upload the
build as a zip file, preserving the template file name. Otherwise,
CodePipeline assigns a random name to the template which is impossible
for the deploy stage to predict.

### DIFF
--- a/configuration/exodus-pipeline.yaml
+++ b/configuration/exodus-pipeline.yaml
@@ -35,15 +35,6 @@ Conditions:
   NotProd:
     !Not [!Equals [!Ref env, prod]]
 
-Mappings:
-  SourceMap:
-    dev:
-      objectKey: ""
-    stage:
-      objectKey: build-dev/exodus-lambda-pkg.yaml
-    prod:
-      objectKey: build-stage/exodus-lambda-pkg.yaml
-
 Resources:
   SNSTopic:
     Type: AWS::SNS::Topic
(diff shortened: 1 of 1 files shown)


configuration/exodus-pipeline.yaml
```diff
@@ -126,10 +117,13 @@ Resources:
                 Configuration:
                   S3Bucket: exodus-pipeline-artifacts
                   S3ObjectKey:
-                    !FindInMap [SourceMap, !Ref env, objectKey]
-                  PollForSourceChanges: false
+                    !If
+                      - NotProd
+                      - build-dev
+                      - build-stage
+                  PollForSourceChanges: true
```
rohanpm, line 124: Do you know whether this PollForSourceChanges is necessary?  If so, what's the poll interval?  I thought this is typically event-driven and not relying on polling.

nathanegillett: I thought so as well but in practice didn't see the event trigger working. I'll give it another try today.

nathanegillett: So, event-driven change detection isn't automatic in our case and will require the creation of a CloudWatch event in the pipeline template.

I'd like to add this instead of relying on polling (the interval for which isn't made clear anywhere) but will submit a follow-up PR with that change.

PR #52, opened 2020-04-25
Enable configuration of exodus-lambda loggers [RHELDST-662]

crungehottman:
The loggers are configured according to a config file in the deployment
package. The levels, formatting, and handlers of the loggers may be
configured in accordance with the logging.config.dictConfig documentation.

By default, the current behavior of the loggers is maintained; the
loggers are configured to log messages at the warning level and higher.

### DIFF
--- a/configuration/lambda_config.json
+++ b/configuration/lambda_config.json
@@ -12,5 +12,27 @@
   },
   "headers": {
     "max_age": "600"
+  },
+  "logging": {
+    "version": 1,
+    "incremental": "True",
+    "disable_existing_loggers": "False",
+    "formatters": {
+      "default": {
+        "format": "%(asctime)s - %(levelname)s - %(aws_request_id)s - %(message)s\n",
+        "datefmt": "%Y-%m-%d %H:%M:%S"
+      }
+    },
+    "loggers": {
+      "origin-response": {
+        "level": "WARNING"
+      },
+      "origin-request": {
+        "level": "WARNING"
+      },
+      "default": {
+        "level": "WARNING"
+      }
+    }
   }
 }
--- a/exodus_lambda/functions/base.py
+++ b/exodus_lambda/functions/base.py
@@ -1,18 +1,24 @@
 import json
 import os
+import logging.config
+import logging
 
 
 class LambdaBase(object):
-    def __init__(self, conf_file="lambda_config.json"):
+    def __init__(self, logger_name="default", conf_file="lambda_config.json"):
         self._conf_file = conf_file
         self._conf = None
+        self._logger_name = logger_name
+        self._logger = None
 
     @property
     def conf(self):
         if not self._conf:
-            with open(self._conf_file, "r") as json_file:
-                self._conf = json.load(json_file)
-
+            if isinstance(self._conf_file, dict):
+                self._conf = self._conf_file
+            else:
+                with open(self._conf_file, "r") as json_file:
+                    self._conf = json.load(json_file)
         return self._conf
 
     @property
@@ -23,6 +29,20 @@ class LambdaBase(object):
         if env_region in self.conf["table"]["available_regions"]:
             return env_region
         return self.conf["table"]["available_regions"][0]
+
+    @property
+    def logger(self):
+        if not self._logger:
+            log_conf = self.conf.get("logging", {})
+            logging.config.dictConfig(log_conf)
+            root_logger = logging.getLogger()
+            if log_conf and root_logger.handlers:
+                formatter_str = log_conf["formatters"]["default"]["format"]
+                formatter_date = log_conf["formatters"]["default"]["datefmt"]
+                formatter = logging.Formatter(formatter_str, formatter_date)
+                root_logger.handlers[0].setFormatter(formatter)
+            self._logger = logging.getLogger(self._logger_name)
+        return self._logger
 
     def handler(self, event, context):
         raise NotImplementedError
--- a/tests/functions/test_base.py
+++ b/tests/functions/test_base.py
@@ -1,8 +1,12 @@
 import pytest
 
+import logging
+
 from exodus_lambda.functions.base import LambdaBase
+from test_utils.utils import generate_test_config
 
 CONF_PATH = "configuration/lambda_config.json"
+TEST_CONF = generate_test_config(CONF_PATH)
 
 
 def test_base_handler():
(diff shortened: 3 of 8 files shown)


exodus_lambda/functions/base.py
```diff
@@ -1,18 +1,32 @@
 import json
+import logging.config
+import logging
 
 
 class LambdaBase(object):
-    def __init__(self, conf_file="lambda_config.json"):
+    def __init__(self, logger_name="default", conf_file="lambda_config.json"):
```
nathanegillett, line 7: Should add a "default" logger to the config

exodus_lambda/functions/base.py
```diff
@@ -1,18 +1,32 @@
...
             with open(self._conf_file, "r") as json_file:
                 self._conf = json.load(json_file)
-
         return self._conf
 
+    @property
+    def logger(self):
+        if not self._logger:
+            logging.config.dictConfig(self.conf.get("logging") or {})
+            if self.conf["logging"]["loggers"][self._logger_name].get(
```
rohanpm, line 24: What's the purpose of this branch?  Why do we want to disable all handlers on the root logger, isn't that going to remove our ability to collect log messages from any loggers other than the ones we listed in config?  Like warnings from one of the Amazon libraries?

configuration/lambda_config.json
```diff
@@ -5,5 +5,35 @@
   },
   "headers": {
     "max_age": "600"
+  },
+  "logging": {
+    "version": 1,
+    "formatters": {
+      "brief": {
+        "format": "%(asctime)s - %(levelname)s - %(message)s",
```
rohanpm, line 13: Is this throwing away some useful info compared to the default handler?  That handler seems to somehow be able to add the lambda request/event ID into the log messages, we'd better keep that if possible.

tests/functions/test_origin_request.py
```diff
@@ -115,13 +122,19 @@ def test_origin_request_invalid_item(
...
 
-    assert (
-        "Exception occurred while processing %s"
-        % json.dumps(
+    error_message = "Exception occurred while processing %s" % json.dumps(
+        {
+            "web_uri": {"S": TEST_PATH},
+            "from_date": {"S": "2020-02-17T00:00:00.000+00:00"},
+        }
+    )
```
nathanegillett, line 130: This isn't being used?

crungehottman: oops, thanks :+1:

exodus_lambda/functions/base.py
```diff
@@ -1,18 +1,35 @@
...
         return self._conf
 
+    @property
+    def logger(self):
+        if not self._logger:
+            log_conf = self.conf.get("logging", {})
+            logging.config.dictConfig(log_conf)
+            root_logger = logging.getLogger()
+            self._logger = logging.getLogger(self._logger_name)
+            if log_conf:
```
rohanpm, line 27:
```suggestion
            if log_conf and root_logger.handlers:
```

In a clean Python environment, the root logger starts with no handlers, so I think this should probably have a check for root_logger.handlers and not proceed if it's empty.

```
$ python3
Python 3.7.5 (default, Dec 15 2019, 17:54:26) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> logging.getLogger().handlers
[]
```

The Lambda runtime seems to always install a handler, but I think it's better for the code not to crash if that didn't happen, mainly for the sake of testing.  If tests right now are not crashing here, I guess that's only because something in the test framework has added a handler on the root logger?  Seeing as I don't know where that's coming from, I'd prefer us not to rely on it.

tests/functions/test_origin_request.py
```diff
@@ -70,7 +71,17 @@ def test_origin_request(
 
     event = {"Records": [{"cf": {"request": {"uri": req_uri, "headers": {}}}}]}
 
-    request = OriginRequest(conf_file=CONF_PATH).handler(event, context=None)
+    request_obj = OriginRequest(conf_file=CONF_PATH)
+    # request.conf["logging"] = logger_conf
```
rohanpm, line 75:
```suggestion
```

tests/functions/test_origin_request.py
```diff
@@ -70,7 +71,17 @@ def test_origin_request(
...
+    request_obj.conf["logging"]["formatters"]["default"][
```
rohanpm, line 76: This is weird, you're adjusting the config after the object is created.

I guess this is relying on the behavior that the logger isn't configured until the first time it's used, and it's not used within the constructor?  This is setting a bit of a trap, e.g. adding a logging statement in \_\_init\_\_ seems like an innocent thing to do, but if a developer tries that then they'll cause this test to start failing.  Ideally the test should not be so brittle.

Can we just feed the desired config into the object when constructing it instead?  Why do we feed one "wrong" config, and then modify it later, relying on the implementation detail that the config hasn't been accessed yet?

exodus_lambda/functions/base.py
crungehottman: Good call. Apparently, if one does not add a `-p no:logging` arg to pytest, pytest adds a `<LogCaptureHandler (NOTSET)>` handler to the root logger for/by the caplog fixture.

crungehottman: > I think this should probably have a check for root_logger.handlers and not proceed if it's empty

@rohanpm , I wasn't sure if by 'not proceed' if you meant to gracefully except an error, or to add a handler.

rohanpm: I just mean don't enter this block, i.e. if there is a handler then configure it, if there is no handler then do nothing.

The primary concern is just to not crash so that we've not added a precondition onto this code, "root logger must have at least one handler to use this class".

exodus_lambda/functions/base.py
```diff
@@ -1,18 +1,37 @@
...
-
         return self._conf
 
+    @property
+    def logger(self):
+        if not self._logger:
+            log_conf = self.conf.get("logging", {})
+            logging.config.dictConfig(log_conf)
+            root_logger = logging.getLogger()
+            if not root_logger.handlers:
```
rohanpm, line 26: This is changing the crash from one kind of exception to another, but in the suggested change I posted, it would avoid the crash.  I'd prefer to do that...

What I'm getting at here is that, even though the lambda runtime is always going to set a handler, making the code crash if that doesn't exist is making the code harder to work with.  For example if you wanted to experiment at the python console or wanted to write a little test script which creates an instance of this class and invokes it with some inputs, then we've made it more difficult to do that right now, as the class now has this hidden dependency on the root logger having a handler.  It only cost a tiny bit of code to avoid that, so I think it's worthwhile.

tests/functions/test_base.py
```diff
@@ -1,10 +1,42 @@
...
+CONF_PATH = "configuration/lambda_config.json"
+MOCKED_CONF = generate_test_config(CONF_PATH)
 
 
 def test_base():
     with pytest.raises(NotImplementedError):
         LambdaBase(conf_file=CONF_PATH).handler(event=None, context=None)
+
+
+@mock.patch("exodus_lambda.functions.base.json.loads")
```
rohanpm, line 18: Could we please avoid this mocking?  This bakes a dependency into many tests that json.loads is used for loading config files (and not for anything else), despite that those tests should have no interest in exactly how the config is loaded.

If the idea is that it's too annoying to write the config dict to a file for load: we control the class under test, so can we just make it accept a dict in the first place?  "A class should be testable using only its public API" is a desirable goal, it's valid to tweak the API a little to serve this purpose.  If the class is cumbersome to test and requires mocking just to initialize it with desired config, then we didn't design it well for our own purposes.

My suggestion is that conf_file should be handled as:

- if it's a dict, then it's the config
- else it's a path to a JSON file

exodus_lambda/functions/base.py
crungehottman: Ah, I understand now. Thanks!

PR #53, opened 2020-04-27
Enable multiple table regions in config, origin_request [RHELDST-663]

nathanegillett:
This commit adds a list of available AWS service endpoints (regions) to
lambda_config.json. The origin_request function will use the first of
the listed regions if an unavailable or invalid region is specified in
the environment variables.

### DIFF
--- a/exodus_lambda/functions/base.py
+++ b/exodus_lambda/functions/base.py
@@ -1,4 +1,5 @@
 import json
+import os
 
 
 class LambdaBase(object):
@@ -13,6 +14,15 @@ class LambdaBase(object):
                 self._conf = json.load(json_file)
 
         return self._conf
+
+    @property
+    def region(self):
+        # Use environment region if among available regions.
+        # Otherwise, default to first available region listed.
+        env_region = os.environ.get("AWS_REGION")
+        if env_region in self.conf["table"]["available_regions"]:
+            return env_region
+        return self.conf["table"]["available_regions"][0]
 
     def handler(self, event, context):
         raise NotImplementedError
--- a/tests/functions/test_base.py
+++ b/tests/functions/test_base.py
@@ -2,9 +2,31 @@ import pytest
 
 from exodus_lambda.functions.base import LambdaBase
 
-CONF_PATH = "exodus_lambda/functions/lambda_config.json"
+CONF_PATH = "configuration/lambda_config.json"
 
 
-def test_base():
+def test_base_handler():
     with pytest.raises(NotImplementedError):
         LambdaBase(conf_file=CONF_PATH).handler(event=None, context=None)
+
+
+@pytest.mark.parametrize(
+    "env_var,exp_var",
+    [
+        ("us-south-7", "us-east-1"),
+        ("ap-southeast-2", "ap-southeast-2"),
+        ("ap-northeast-3", "us-east-1"),
+    ],
+    ids=["fake", "available", "unavailable"],
+)
+def test_base_region(env_var, exp_var, monkeypatch):
+    """Ensure correct regions are selected for various inputs"""
+    base = LambdaBase(conf_file=CONF_PATH)
+
+    # Environment variable is set
+    monkeypatch.setenv("AWS_REGION", env_var)
+    assert base.region == exp_var
+
+    # Environment variable is unset
+    monkeypatch.delenv("AWS_REGION")
+    assert base.region == "us-east-1"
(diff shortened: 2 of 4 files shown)


exodus_lambda/functions/base.py
```diff
@@ -14,5 +14,14 @@ def conf(self):
 
         return self._conf
 
+    @property
+    def region(self):
+        # Use specified region if among available regions.
+        # Otherwise, default to first available region listed.
+        table_conf = self.conf["table"]
+        if table_conf["region"] in table_conf["available_regions"]:
```
rohanpm, line 22: I don't think this is correct:

It's looking at config["table"]["region"].  The value in the config file template in the repo is $AWS_REGION.  We expect that value to be replaced with a real region while building the deployment package for our lambdas, right?

However, there is only *one* deployment package which is used to run the lambda in *multiple* regions so it can't possibly be correct for all regions.

I think "region" should disappear from the config file entirely and should be replaced with os.environ.get("AWS_REGION").

nathanegillett: We need the config file to handle any environment variables, as Lambda@Edge won't accept them.

rohanpm: I don't think that's right.  I think the restriction on Lambda@Edge is that you can't define your own environment variables in the typical way as done for lambdas, but AWS_REGION and maybe some other built-in env vars provided by the platform are still usable.  I've dumped os.environ in a lambda before and it contained AWS_REGION.

In the issue description, I also linked blog post https://aws.amazon.com/blogs/networking-and-content-delivery/leveraging-external-data-in-lambdaedge/ which shows using AWS_REGION environment variable from a lambda.

nathanegillett: That's right. I hadn't considered Lambda having built-in environment variables.

tests/functions/test_base.py
```diff
@@ -1,10 +1,29 @@
...
+    "region",
+    ["us-south-7", "us-east-1", "ap-northeast-3"],
+    ids=["fake", "available", "unavailable"],
+)
+def test_base_region(region):
+    """Ensure correct regions are selected for various inputs"""
+    with mock.patch.dict(os.environ, {"AWS_REGION": region}):
+        base = LambdaBase(conf_file=CONF_PATH)
+        # Ensure region defaults when environment variable is set
+        assert base.region == "us-east-1"
```
rohanpm, line 26: Because us-east-1 is the default region, it seems like there is no test which verifies that the env var actually does anything.  This is the only test being added, and it seems like if you delete the code which reads the env var or change it to read from the wrong env var, then this test would still pass... or did I miss something?

Could we have at least one test which verifies that the value from AWS_REGION is actually used where possible?

tests/functions/test_base.py
```diff
@@ -1,10 +1,33 @@
...
+    [
+        ("us-south-7", "us-east-1"),
+        ("ap-southeast-2", "ap-southeast-2"),
+        ("ap-northeast-3", "us-east-1"),
+    ],
+    ids=["fake", "available", "unavailable"],
+)
+def test_base_region(env_var, exp_var):
+    """Ensure correct regions are selected for various inputs"""
+    with mock.patch.dict(os.environ, {"AWS_REGION": env_var}):
```
rohanpm, line 27: Doesn't really matter here but I want to point out that there is a `monkeypatch` fixture which in many cases is a better way to temporarily override env vars.  https://docs.pytest.org/en/latest/monkeypatch.html

```monkeypatch.setenv("AWS_REGION", env_var)```

The reason why it would usually be better is that it keeps the rest of the environment as it is, while here you're wiping out all the rest of the environment, so AWS_REGION is the only present variable which is quite unrealistic.  But I don't think it will have an impact on this particular test.

Comment (nathanegillett) [reply tests/functions/test_base.py]: Fixed

**Comment by nathanegillett** [comment]
Replaced `with mock.patch.dict` with `monkeypatch.setenv` and `.delenv` in test_base.test_base_region

File: tests/functions/test_base.py [hunk 4]
```diff
@@ -2,9 +2,31 @@
...
+    """Ensure correct regions are selected for various inputs"""
+    base = LambdaBase(conf_file=CONF_PATH)
+
+    # Environment variable is set
+    monkeypatch.setenv("AWS_REGION", env_var)
+    assert base.region == exp_var
+
+    # Environment variable is unset
+    monkeypatch.delenv("AWS_REGION")
+    assert base.region == "us-east-1"
```
nathanegillett, line 32: Replaced `mock.patch.dict` use with `monkeypatch`

PR #54, opened 2020-04-28
Switch to event-based stage, prod pipeline trigger

nathanegillett:
This commit introduces a CloudWatch event, CloudTrail trail, and
CloudWatch service role to enable event-based triggering of stage and
prod pipelines. The change allows us to disable polling of the s3
source.


nathanegillett:
Any objections or suggestions to this, @crungehottman @dichen16?

PR #55, opened 2020-04-30
Implement GPG verification [RHELDST-687]

nathanegillett:
This commit introduces a new lambda function to be deployed manually.
The function is invoked by the pipeline following GitHub source stages,
receiving the git commit ID from the pipeline, and verifies that the
pushed commit has a tag and that the tag is signed by an authorized
entity.


exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,64 @@
...
+
+    try:
+        # 1) Download and import whitelisted GPG keys
+        s3_client.download_file(bucket_name, whitelist_name)
+        subprocess.run(
+            ["gpg", "--import", whitelist_name], check=True, text=True
+        )
+
+        # 2) Clone exodus-lambda and checkout deploy branch
+        repo_url = "https://github.com/release-engineering/exodus-lambda.git"
```
rohanpm, line 37: Do we really need to clone it again... don't we already have a SourceArtifact with the repo?  However I admittedly don't know whether that SourceArtifact zip file includes the .git directory or only the working tree.  If it doesn't include .git then indeed we'd have to clone.

One of the downsides of cloning it again is that you make it harder to test locally and harder to point it at some other repo (e.g. remembering Caleigh's PR where she was hacking it to point at her fork).

exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,64 @@
...
+        subprocess.run(
+            ["git", "clone", repo_url, "-b", "deploy"], check=True, text=True
```
rohanpm, line 39: Why are we passing text=True - we aren't reading stdout or stderr from the process, so does that do anything here?

exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,64 @@
...
+        git_tag = complete_process.stdout
+
+        # 4) Verify tag
+        subprocess.run(["git", "verify-tag", git_tag], check=True, text=True)
+
+        # 5) Update whitelist
+        s3_client.upload_file(whitelist_name, bucket_name)
+
+        # Report success to pipeline
+        pipeline_client.put_job_success_result(jobId=job_id)
```
rohanpm, line 59: This made me wonder: do we really need a special pipeline client object, why can't we just return a value or raise if there's a problem?

But according to https://docs.aws.amazon.com/codepipeline/latest/userguide/action-reference-Lambda.html this is indeed required: "As part of the implementation of the Lambda function, there must be a call to either the PutJobSuccessResult API or PutJobFailureResult API."

That's annoying, I wish they had made it so that the function could simply return a value to indicate success.  It's probably designed that way so you can post the result asynchronously after the lambda function has returned.

exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,64 @@
...
+
+    bucket_name = "exodus-pipeline-artifacts"
+    whitelist_name = ".whitelist.gpg"
+
+    s3_client = boto3.client("s3")
+    pipeline_client = boto3.client("codepipeline")
+
+    try:
+        # 1) Download and import whitelisted GPG keys
+        s3_client.download_file(bucket_name, whitelist_name)
```
rohanpm, line 31: I think it may be better to drop this whitelist download/re-upload stuff and just embed the keys into the pipeline.  For example, included in the UserParameter?  YAML has good support for multi-line strings so it is not going to be too ugly compared to the .whitelist.gpg file.

Right now it seems to be like this:

- The pipeline is not self-updating, if you want to change the steps in the pipeline or their config, those will not be updated simply by making the change in "deploy" branch.  You need to explicitly arrange an update to the pipeline.
- An exception is made for the GPG key whitelist, we wrote extra code to retrieve and store the whitelist during deployment so that this part of the pipeline's config can be self-updating.

I have to wonder: why do we need that exception?  It's not like this will need to be updated frequently.

Thinking of it another way.  If I wanted to change, say, the git branch used by the pipeline or the mailing list used for notifications, the process would be: update the YAML file, then manually run some documented command(s) to redeploy the pipeline with the updated config.  Right?

Is there any need for the process for updates to the GPG key list to be any different?  Isn't it just another piece of config for the pipeline, like the repo URL/branch or mailing list, which could be updated in the same way.

exodus_lambda/functions/verify_source.py
nathanegillett: Yeah, there are a lot of small inconveniences with AWS. The put job success call also allows provides a way to pass variables back to the pipeline, which could be useful in some cases.

exodus_lambda/functions/verify_source.py
nathanegillett: Nope. I'll remove it.

exodus_lambda/functions/verify_source.py
nathanegillett: Yeah, like the previous attempt at GPG verification in the buildspec, we have to clone the repo to access the git metadata, as it isn't included in the source zip sent from GitHub.

configuration/exodus-pipeline.yaml
```diff
@@ -125,6 +126,21 @@ Resources:
...
+              - Name: Verify
+                ActionTypeId:
+                  Category: Invoke
+                  Owner: AWS
+                  Provider: Lambda
+                  Version: 1
+                Region: !Ref region
+                Configuration:
+                  FunctionName: verify_source
+                  UserParameters: "#{SourceVars.CommitId}"
```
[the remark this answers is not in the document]
nathanegillett: Yeah, I can add this.

exodus_lambda/functions/verify_source.py
nathanegillett: That makes sense. I'll put pub keys directly into the invoke action's UserParameters.

exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    pipeline_client = boto3.client("codepipeline")
+
+    try:
+        # Import whitelisted GPG keys
+        pub_keys = "".join(
+            pub_key for _, pub_key in job_params["whitelist"].items()
+        )
+        subprocess.run(["gpg", "--import", pub_keys], check=True)
```
rohanpm, line 31: Hmm, sorry to ask but does it work?  If I understand right, pub_keys is going to be a string like 

```
-----BEGIN PGP PUBLIC KEY BLOCK-----
... [key1]
-----END PGP PUBLIC KEY BLOCK-----
-----BEGIN PGP PUBLIC KEY BLOCK-----
... [key2]
-----END PGP PUBLIC KEY BLOCK-----
```

I am surprised that gpg would accept this entire string as a command-line argument.  I would have expected you need to feed it to stdin, or save it to a file first and point gpg at the file, or similar.  The argument after --import is usually interpreted as a filename.

exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,56 @@
...
+    job_data = event["CodePipeline.job"]["data"]
+    job_params = json.loads(
+        job_data["actionConfiguration"]["configuration"]["UserParameters"]
+    )
+
+    pipeline_client = boto3.client("codepipeline")
+
+    try:
+        # Import whitelisted GPG keys
+        pub_keys = "".join(
```
rohanpm, line 28: Is "".join right or does it have to be "\n".join ?

In the YAML file for example you have:

```
                                   -----END PGP PUBLIC KEY BLOCK-----",
                      "ngillett": "-----BEGIN PGP PUBLIC KEY BLOCK-----
```

Note the keys don't end with a newline.  If you joined those items then wouldn't you end up with a line like:

```
-----END PGP PUBLIC KEY BLOCK----------BEGIN PGP PUBLIC KEY BLOCK-----
```

Not sure if the format is documented but I'd have assumed that newlines were significant at least around the key begin/end delimiters.

exodus_lambda/functions/verify_source.py
rohanpm: In that case, if it's not much trouble do you think you could add the URL and branch into the pipeline and pass them in as UserParameters as well?  Or: is it possible to extract them from the Source action's output variables?  According to https://docs.aws.amazon.com/codepipeline/latest/userguide/action-reference-GitHub.html#action-reference-GitHub-variables , it will export RepositoryName and BranchName, so it seems like it should be possible to reuse them rather than hardcoding here.  Or maybe they are even in job_params already like the commit_id is?

It would be nice to avoid someone having to hack this to point at their own repo if they want to test from their own fork.

exodus_lambda/functions/verify_source.py
```diff
@@ -0,0 +1,56 @@
...
+            pub_key for _, pub_key in job_params["whitelist"].items()
+        )
+        subprocess.run(["gpg", "--import", pub_keys], check=True)
+
+        # Clone exodus-lambda and checkout deploy branch
+        repo_url = "https://github.com/release-engineering/exodus-lambda.git"
+        subprocess.run(["git", "clone", repo_url, "-b", "deploy"], check=True)
+        os.chdir("exodus-lambda")
```
rohanpm, line 36: Just a quick note that if you do start accepting the repo name as a parameter, this part would ideally not be hardcoded any more.

I would suggest using the form of git clone where you specify the output directory instead of letting git choose for you, e.g. "git clone <repo_url> src -b deploy" and then you know it's always cloned to "src" regardless of the repo's name.

exodus_lambda/functions/verify_source.py
nathanegillett: Agreed.
The branch and repo names from the source stage variables are usable. Though, one would still need to change owner, repo and branch values in source stage to point the pipeline elsewhere. So, I'll just create owner, repo and branch template parameters and supply them to source and verify stages.

exodus_lambda/functions/verify_source.py
nathanegillett: I've confirmed that they needs to be stored in a file.

rohanpm:
We discussed in a call that you think this approach won't work any more and you're investigating a buildspec-based approach instead which seems promising.  In that case, I think that means this doesn't need review any more.  Please correct me if I'm wrong.

nathanegillett:
Yeah, I suppose I can create a new PR since this commit is basically being erased.

PR #56, opened 2020-05-06
Implement GPG verification [RHELDST-687]

nathanegillett:
This commit introduces a new CodeBuild project to be deployed manually
via CloudFormation prior to the creation of CodePipelines. The project
is run by the pipeline following GitHub source stages, receiving the
git commit ID from the pipeline, and verifies that the pushed commit
has a tag that is signed by an authorized entity.

This verify-source CodeBuild project requires the following environment
variables;
- REPO_OWNER: Name of the repository's owner
- REPO_NAME: Name of the repository
- REPO_BRANCH: Name of the repository's branch
- COMMIT_ID: ID of the commit to verify


configuration/verify-source.yaml
```diff
@@ -0,0 +1,57 @@
...
+        Type: NO_SOURCE
+        BuildSpec:
+          !Sub
+          "version: 0.2\n\nphases:\n  build:\n    commands:\n      -
+           aws s3 cp s3://exodus-pipeline-artifacts/.whitelist.gpg .whitelist.gpg\n      -
+           gpg --import .whitelist.gpg\n      -
+           git clone https://github.com/$REPO_OWNER/$REPO_NAME.git -b $REPO_BRANCH\n      -
+           cd ./$REPO_NAME\n      -
+           git verify-tag $(git describe --exact-match $COMMIT_ID)\n      -
+           aws s3 cp .whitelist.gpg s3://exodus-pipeline-artifacts/.whitelist.gpg"
```
nathanegillett, line 25: As done in a previous attempt at tag verification, this downloads the whitelist from s3 and, once the commit is verified, updates it with the version provided by the commit (reminder that versioning is enabled in the bucket for history).

We discussed before that, instead of doing this, hard-coding the GPG keys was a better option. However, this doesn't work well here, if at all. I think that, if we were to force it, it would be far too difficult to manage, as very specific formatting/use of control characters would likely be necessary.

nathanegillett:
Build using this spec with environment variables pointing to this branch (687) can be seen here: https://console.aws.amazon.com/codesuite/codebuild/projects/verify-source/build/verify-source%3A117f297f-8a11-42b1-b932-23165ed51c36/log?region=us-east-1

configuration/verify-source.yaml
```diff
@@ -0,0 +1,57 @@
...
+      Name: verify-source
+      ServiceRole: !GetAtt CodeBuildRole.Arn
+      Environment:
+        Type: LINUX_CONTAINER
+        ComputeType: BUILD_GENERAL1_SMALL
+        Image: aws/codebuild/amazonlinux2-x86_64-standard:2.0
+      Source:
+        Type: NO_SOURCE
+        BuildSpec:
+          !Sub
```
rohanpm, line 18: Is this !Sub needed?  What's being substituted exactly?  It doesn't seem like it could be the env vars.  e.g. $COMMIT_ID could not even possibly be known at the point the !Sub is evaluated.

If you remove it, you ought to be able to use a nicer form of YAML multi-line string.  Does that change the situation with regards to embedding the GPG keys?

configuration/exodus-pipeline.yaml
```diff
@@ -130,6 +143,25 @@ Resources:
...
+                ActionTypeId:
+                  Category: Build
+                  Owner: AWS
+                  Provider: CodeBuild
+                  Version: 1
+                Region: !Ref region
+                Configuration:
+                  EnvironmentVariables:
+                    !Sub '[
+                      {"name": "REPO_OWNER", "value": ${repoOwner}, "type": "PLAINTEXT"},
```
rohanpm, line 156: I notice that the quoting isn't consistent between this place and a !Sub later on in the change.  e.g. you write ${repoOwner} here but write "${env}" in the other one.

I don't have a good understanding of how exactly the string in !Sub is evaluated, but if you tested this and it works then I guess neither one of them needs the quoting... which would make sense if this is being evaluated as YAML.

configuration/verify-source.yaml
nathanegillett: Nope, that sub function was left over from experimentation. Thanks.

The YAML string looks better now but I don't think the public key issue is helped at all. The whitelist string would still need to be formatted very specifically in order for gpg to recognize it.

configuration/exodus-pipeline.yaml
nathanegillett: This was a mistake; the value should be wrapped in double-quotes.

The sub function requires the ${xxxx} syntax but EnvironmentVariables requires a JSON format string.

You'll also see #{xxxx} a few times, this syntax is used for pipeline variables.
https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-variables.html

nathanegillett:
Build results with this spec: https://console.aws.amazon.com/codesuite/codebuild/projects/verify-source/build/verify-source%3A780bf59c-da1d-4324-95df-ab4a135c617c/log?region=us-east-1

configuration/verify-source.yaml
```diff
@@ -0,0 +1,59 @@
...
+        Image: aws/codebuild/amazonlinux2-x86_64-standard:2.0
+      Source:
+        Type: NO_SOURCE
+        BuildSpec: |
+          version: 0.2
+          phases:
+            build:
+              commands:
+                - aws s3 cp s3://exodus-pipeline-artifacts/.whitelist.gpg .whitelist.gpg
+                - gpg --import .whitelist.gpg
```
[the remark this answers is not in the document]
nathanegillett: This doesn't work, as `gpg --import` can't open a here-document.
https://console.aws.amazon.com/codesuite/codebuild/projects/verify-source/build/verify-source%3Adb0e2a45-109d-45e2-b457-59297f42caaf/log?region=us-east-1

So, first writing the keys to a file (`echo '...' >> .whitelist.gpg`) _should_ fix things. This doesn't work either, unfortunately.
https://console.aws.amazon.com/codesuite/codebuild/projects/verify-source/build/verify-source%3A7c50d9fd-dd4d-4d38-85f9-43a925513c93/log?region=us-east-1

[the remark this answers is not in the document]
nathanegillett: Also, yes, using this method means an initial whitelist needs to be uploaded to the pipeline bucket.

PR #57, opened 2020-05-12
[RHELDST-590] Create reference testdata for Exodus

dichen16:
The test data specifying in the data.yml is not complete, I will create another PR for collecting test data.
I am not sure if using library `tqdm` to visualize the download progress is needed.
```
# prepare
(cdn_lambda_env) [dichen@dilaptop autotests]$ ./reftest.py prepare --cert ~/certs/cdn_debug_cert/rcm-debug.crt --key ~/certs/cdn_debug_cert/rcm-debug.key --cacert /etc/rhsm/ca/redhat-uep.pemhttps://cdn.redhat.com/content/dist/rhel/server/5/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████████| 30.0/30.0 [00:00<00:00, 20.9kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/5.7/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████| 17.0/17.0 [00:00<00:00, 13.0kiB/s]
https://cdn.redhat.com/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm: 100%|████████████████████████████████████████████| 196k/196k [00:05<00:00, 39.0kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/Packages/c/cloud-init-0.7.9-9.el7.x86_64.rpm: 100%|████████████████████████████████████████| 636k/636k [00:01<00:00, 508kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/repomd.xml: 3.53kiB [00:00, 2.31MiB/s]                                                                               
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2: 100%|███████████████| 12.0M/12.0M [00:09<00:00, 1.24MiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz: 100%|███████████████| 25.9M/25.9M [00:05<00:00, 5.08MiB/s]
# withdraw
(cdn_lambda_env) [dichen@dilaptop reftest]$ ./reftest.py clean
```

### DIFF
--- a/support/reftest/data.yml
+++ b/support/reftest/data.yml
@@ -1,0 +1,25 @@
+prod-cdn-url:
+  https://cdn.redhat.com
+
+test_data:
+- path: /content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2
+  sha256: 647b38c2222caecd0859f97a0ef0ef78f249b8d96af437552d6b2950056f261e
+  content-type: application/x-bzip
+- path: /content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz
+  sha256: 49159602b5bb8d9d34af384a62bd411d12449338501a4d08ef90c369b7facb98
+  content-type: application/x-gzip
+# rpm
+- path: /content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm
+  sha256: 06316538b90d3aab20f4787132307f6d330da9a48bab11a13e0a616fcf622ce5
+  content-type: audio/x-pn-realaudio-plugin
+# listing
+- path: /content/dist/rhel/server/5/5.7/listing
+  sha256: b51f4ddc06fddec9e73892671f1f25300ccd4235fa798afd01caf96446dc2bf1
+  content-type: text/plain
+- path: /content/dist/rhel/server/5/5.6/listing
+  sha256: b51f4ddc06fddec9e73892671f1f25300ccd4235fa798afd01caf96446dc2bf1
+  content-type: text/plain
+- path: /content/dist/rhel/server/5/listing
+  sha256: 6e0230c69d88400d6064a4fce264aebb6918da0f12c86790c2bfc6edd3851907
+  content-type: text/plain
+
--- a/support/reftest/reftest
+++ b/support/reftest/reftest
@@ -1,0 +1,246 @@
+#!/usr/bin/env python
+"""
+
+Prepare reference test data for Exodus-lambda integration test.
+It requires a "data.yml" file specifying the reference test data.
+
+"""
+
+import os
+import sys
+import yaml
+import boto3
+import hashlib
+import tempfile
+import requests
+import argparse
+
+from tqdm import tqdm
+from datetime import datetime
+
+
+class DBHandler:
+    def __init__(self, table, session):
+        self.dynamodb = session.client("dynamodb")
+        self.table = table
+
+        self.default_from_date = datetime.utcnow().isoformat(
+            timespec="seconds"
+        )
+
+    def put_item(self, web_uri, object_key, from_date=None, metadata={}):
+        if not from_date:
+            from_date = self.default_from_date
+        return self.dynamodb.put_item(
+            TableName=self.table,
+            Item={
+                "web_uri": {"S": web_uri},
+                "from_date": {"S": from_date},
+                "object_key": {"S": object_key},
+                "metadata": {"M": metadata},
+            },
+        )
+
+
+class S3Handler:
+    def __init__(self, bucket, session):
+        self.s3 = session.client("s3")
+        self.bucket = bucket
+
+    def upload_from_localfile(self, path, checksum, str_content_type):
+        if not str_content_type:
+            dict_content_type = {}
+        else:
+            dict_content_type = {"ContentType": str_content_type}
+
+        with open(path, "rb") as data:
+            self.s3.upload_fileobj(
+                Fileobj=data,
+                Bucket=self.bucket,
+                Key=checksum,
+                ExtraArgs=dict_content_type,
+            )
+
+
+def parse_aws_session(parser):
+    parser.add_argument(
+        "--aws-access-id",
+        default=None,
+        help="Access ID for Amazon services. If no ID is provided, attempts to"
+        " find it among environment variables and ~/.aws/config file will"
+        " be made",
+    )
+    parser.add_argument(
+        "--aws-access-key",
+        default=None,
+        help="Access key for Amazon services. If no key is provided, attempts"
+        " to find it among environment variables and ~/.aws/config file"
+        " will be made",
+    )
+    parser.add_argument(
+        "--aws-session-token",
+        default=None,
+        help="Session token for Amazon services. If no token is provided,"
+        " attempts to find it among environment variables and"
+        " ~/.aws/config file will be made",
+    )
+    parser.add_argument(
+        "--default-region",
+        default="us-east-1",
+        help="Default region for Amamzon services. If no region is provided,"
+        " it will go with us-east-1",
+    )
+
+
+def parse_cdn_certs(parser):
+    parser.add_argument(
+        "--cert",
+        default="~/certs/rcm-debug/rcm-debug.crt",
+        type=str,
+        help="Client certificate file and password",
+    )
+    parser.add_argument(
+        "--key",
+        default="~/certs/rcm-debug/rcm-debug.key",
+        type=str,
+        help="Private key file name",
+    )
+    parser.add_argument(
+        "--cacert",
+        default="/etc/rhsm/ca/redhat-uep.pem",
+        type=str,
+        help="CA certificate to verify peer against",
+    )
+
+
+def parse_args():
+    root_parser = argparse.ArgumentParser(description=__doc__)
+    subparsers = root_parser.add_subparsers(
+        dest="command", help="reference test data operations"
+    )
+    # parser for prepare operation
+    parser_prepare = subparsers.add_parser(
+        "prepare",
+        help="""
+        prepare reference test data in dynamodb and s3
+        for integration test,
+         e.g. $./reftest prepare --bucket exodus-bucket --table exodus-table
+        """,
+    )
+    parser_prepare.add_argument(
+        "--release-date",
+        default=None,
+        help="Date on which the content will be made available.",
+    )
+    parse_aws_session(parser_prepare)
+    parse_cdn_certs(parser_prepare)
+    parser_prepare.add_argument(
+        "--bucket",
+        required=True,
+        help="The AWS S3 bucket used to store test data",
+    )
+    parser_prepare.add_argument(
+        "--table",
+        required=True,
+        help="The AWS dynamoDB used to store test data",
+    )
+
+    return root_parser.parse_args()
+
+
+class RefTestConfig:
+    def __init__(self, prod_cdn_url, test_data):
+        self.prod_cdn_url = prod_cdn_url
+        self.test_data = test_data
+
+
+def load_config():
+    with open("data.yml") as f:
+        config = yaml.load(f, yaml.SafeLoader)
+    return RefTestConfig(config["prod-cdn-url"], config["test_data"],)
+
+
+# It will return a TempFileObj and a checksum for test data verification
+def download_to_local(url, key_path, cert_path, cacert_path):
+    with requests.get(
+        url,
+        cert=(cert_path, key_path),
+        verify=cacert_path,
+        stream=True,
+        timeout=(30, 30),
+    ) as r:
+        r.raise_for_status()
+        total_size = int(r.headers.get("content-length", 0))
+        t = tqdm(desc=url, total=total_size, unit="iB", unit_scale=True)
+
+        # the file is deleted as soon as it is closed.
+        temp_file = tempfile.NamedTemporaryFile(delete=True)
+
+        sha256 = hashlib.sha256()
+
+        for chunk in r.iter_content(chunk_size=8192):
+            if chunk:  # filter out keep-alive new chunks
+                t.update(len(chunk))
+                temp_file.write(chunk)
+                sha256.update(chunk)
+                temp_file.flush()
+        t.close()
+        return temp_file, sha256.hexdigest()
+
+
+def prepare(db, s3, config, opt):
+    for item in config.test_data:
+        url = config.prod_cdn_url + item["path"]
+        # download test data to local with a name "NamedTemporaryFile"
+        temp_file, cdn_data_checksum = download_to_local(
+            url, opt.key, opt.cert, opt.cacert
+        )
+
+        # checksum verify
+        if cdn_data_checksum != item["sha256"]:
+            print(
+                "{} verify checksum failed, cdn_data_checksum is {}, but test_data_checksum is {}".format(
+                    item["path"], cdn_data_checksum, item["sha256"]
+                )
+            )
+            return False
+
+        # push test data to s3 and dynamodb
+        s3.upload_from_localfile(
+            temp_file.name, item["sha256"], item["content-type"]
+        )
+        db.put_item(item["path"], item["sha256"], opt.release_date)
+
+        # delete the NamedTemporaryFile
+        temp_file.close()
+
+    return True
+
+
+def main():
+    config = load_config()
+    opt = parse_args()
+
+    session = boto3.Session(
+        aws_access_key_id=opt.aws_access_id,
+        aws_secret_access_key=opt.aws_access_key,
+        aws_session_token=opt.aws_session_token,
+        region_name=opt.default_region,
+    )
+
+    db_client = DBHandler(table=opt.table, session=session)
+    s3_client = S3Handler(bucket=opt.bucket, session=session)
+
+    res = False
+    if opt.command == "prepare":
+        res = prepare(db_client, s3_client, config, opt)
+
+    if res:
+        print("{} operation has finished successfully!".format(opt.command))
+    else:
+        print("Fatal error: {} operation is terminated".format(opt.command))
+        return 1
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,3 @@
+tqdm
+pyyaml
+requests


nathanegillett:
Looks all right to me. I'd like to see a body on the commit message, though. And some more documentation in the script might be helpful.

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+            TableName=self.table,
+            Key={
+                "web_uri": {"S": res["Items"][0]["web_uri"]["S"]},
+                "from_date": {"S": res["Items"][0]["from_date"]["S"]},
+            },
+        )
+
+    def put_item(self, web_uri, object_key, from_date=None, metadata={}):
+        if not from_date:
+            from_date = datetime.utcnow().isoformat(timespec="seconds")
```
rohanpm, line 74: Why does this one use timespec="seconds" while the searching code uses timespec="milliseconds" ?

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+        "--default-region",
+        default=None,
+        help="Default region for Amamzon services. If no region is provided,"
+        " attempts to find it among environment variables and"
+        " ~/.aws/config file will be made",
+    )
+
+
+def parse_cdn_certs(parser):
+    parser.add_argument(
```
rohanpm, line 177: For these arguments, there are some existing conventions which could be used for default values.  It would make the command a little easier to run.  Would you mind adding those defaults?

They are:

- cert: ~/certs/rcm-debug/rcm-debug.crt
- key: ~/certs/rcm-debug/rcm-debug.key
- cacert: /etc/rhsm/ca/redhat-uep.pem

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+
+def parse_args():
+    root_parser = argparse.ArgumentParser(description=__doc__)
+    subparsers = root_parser.add_subparsers(
+        dest="command", help="reference test data operations"
+    )
+    parser_prepare = subparsers.add_parser(
+        "prepare",
+        help="""
+        prepare referece test data for intagration test
```
rohanpm, line 197:
```suggestion
        prepare reference test data for integration test
```

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+        default=None,
+        help="Date on which the content will be made available.",
+    )
+    parse_aws_info(parser_prepare)
+    parse_cdn_certs(parser_prepare)
+
+    parser_clean = subparsers.add_parser(
+        "clean",
+        help="""
+        clean referece test data in s3 and dynamodb
```
rohanpm, line 212:
```suggestion
        clean reference test data in s3 and dynamodb
```

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+        ).client("dynamodb")
+
+        self.table = table
+
+    def search_item_from_web_uri(self, web_uri):
+        return self.dynamodb.query(
+            TableName=self.table,
+            Limit=1,
+            ScanIndexForward=False,
+            KeyConditionExpression="web_uri = :u and from_date <= :d",
```
rohanpm, line 44: Should we really have the from_date <= check here?  This is used during "clean", right?

To me, it doesn't really make sense right now.  You made it possible for "prepare" to upload the data with a date in the future, but made it impossible for "clean" to clean that data.  Shouldn't they be consistent?  And why should the semantics of "clean" be only to delete the most recent non-future entry for any given URL?  (Which is not really "clean", it's "roll back to previous version").

I'm not quite understanding what is the intended behavior of "clean" and when I'd want to use that.

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+    return RefTestConfig(
+        config["prod-cdn-url"],
+        config["path"],
+        config["destination"]["table"],
+        config["destination"]["bucket"],
+    )
+
+
+def download_to_local(url, key_path, cert_path, cacert_path):
+    local_filename = url.split("/")[-1]
```
rohanpm, line 241: This seems rather dangerous.  You're downloading into the current working directory and overwriting whatever's already there with the same filename.  So, for instance, if I have a file named "listing" in the current directory when I run this command, it's going to be silently deleted.

Should this use a temporary file instead?  (e.g. NamedTemporaryFile)

support/reftest/data.yml
```diff
@@ -0,0 +1,19 @@
...
+path:
+  # listing file
+  - /content/dist/rhel/server/5/listing
+  - /content/dist/rhel/server/5/5.7/listing
+  # packages
+  - /content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm
+  - /content/dist/rhel/server/7/7.4/x86_64/os/Packages/c/cloud-init-0.7.9-9.el7.x86_64.rpm
+  # repodata
+  - /content/dist/rhel/server/7/7.4/x86_64/os/repodata/repomd.xml
+  - /content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2
```
rohanpm, line 17: To elaborate on what I said elsewhere about storing sums and content types in the testdata rather than calculating dynamically.  I think this would be better defined as something like this:

```
- path: /content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2
  sha256: 647b38c2222caecd0859f97a0ef0ef78f249b8d96af437552d6b2950056f261e
  content-type: application/x-bzip
- path: /content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz
  sha256: 49159602b5bb8d9d34af384a62bd411d12449338501a4d08ef90c369b7facb98
  content-type: application/x-gzip
```

So the reference testdata actually defines the most relevant properties of the test data right there, rather than calculating it on the fly with code which could be buggy and would need to be maintained in sync with publishing tool code.

support/reftest/reftest.py
dichen16: > This seems rather dangerous. You're downloading into the current working directory and overwriting whatever's already there with the same filename. So, for instance, if I have a file named "listing" in the current directory when I run this command, it's going to be silently deleted.
> 
> Should this use a temporary file instead? (e.g. NamedTemporaryFile)

@rohanpm solved.

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
...
+import os
+import sys
+import yaml
+import boto3
+
+import hashlib
+import requests
+import argparse
+
+from tqdm import tqdm
```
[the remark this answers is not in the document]
dichen16: @rohanpm requirements-dev.txt added.

nathanegillett:
@dichen16, the commit message isn't very clear and the subject line seems to break into the body.
Would be nice to explain the changes a bit, what solutions or benefits they provide.
Someone looking at the git log isn't going to be able to tell what this commit did, as the message doesn't even specify what files were changed.

support/reftest/reftest
```diff
@@ -0,0 +1,293 @@
...
+        url,
+        cert=(cert_path, key_path),
+        verify=cacert_path,
+        stream=True,
+        timeout=(30, 30),
+    ) as r:
+        r.raise_for_status()
+        total_size = int(r.headers.get("content-length", 0))
+        t = tqdm(desc=url, total=total_size, unit="iB", unit_scale=True)
+        with open("NamedTemporaryFile", "wb") as f:
```
nathanegillett, line 233: When @rohanpm suggested using a NamedTemporaryFile for downloaded files, he was referring to tempfile.NamedTemporaryFile.
https://docs.python.org/3.7/library/tempfile.html#tempfile.NamedTemporaryFile

support/reftest/reftest.py
```diff
@@ -0,0 +1,310 @@
+#!/usr/bin/env python3
```
nathanegillett, line 1: This file will need to be deleted now that we have /reftest, yeah?

dichen16:
> @dichen16, the commit message isn't very clear and the subject line seems to break into the body.
> Would be nice to explain the changes a bit, what solutions or benefits they provide.
> Someone looking at the git log isn't going to be able to tell what this commit did, as the message doesn't even specify what files were changed.

Yes, this commit message is messy, I will create a clean PR

support/reftest/reftest.py
dichen16: @nathanegillett  deleted.

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+
+    def delete_item_from_web_uri(self, web_uri):
+        res = self.search_item_from_web_uri(web_uri)
+        return self.dynamodb.delete_item(
+            TableName=self.table,
+            Key={
+                "web_uri": {"S": res["Items"][0]["web_uri"]["S"]},
+                "from_date": {"S": res["Items"][0]["from_date"]["S"]},
+            },
+        )
```
nathanegillett, line 62: How can we be sure that the table item we want to delete is first in the list of items returned by the query?
Have you considered calculating the checksum of the file downloaded from the CDN and including it in the query (object_key)?

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+            KeyConditionExpression="web_uri = :u",
+            ExpressionAttributeValues={":u": {"S": web_uri}},
+        )
+
+    def search_object_key_from_web_uri(self, web_uri):
+        res = self.search_item_from_web_uri(web_uri)
+        if not res["Items"]:
+            return False
+        else:
+            return res["Items"][0]["object_key"]["S"]
```
nathanegillett, line 52: This isn't guaranteed to be accurate, as the web_uri can be shared by multiple items. So, while this returns **an** object_key, we can't say it's the same file that was downloaded from the CDN.

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+        access_key=None,
+        session_token=None,
+        region_name=None,
+    ):
+        self.dynamodb = boto3.Session(
+            aws_access_key_id=access_id,
+            aws_secret_access_key=access_key,
+            aws_session_token=session_token,
+            region_name=region_name,
+        ).client("dynamodb")
```
nathanegillett, line 34: Since both this and the s3 client are always used, it would probably be better to only create the session once and then create the client in each *Handler class.
So, instead of passing access_id, access_key, session_token, and region_name to each handler, create the session in main() and pass that object to the handlers.

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+        help="Session token for Amazon services. If no token is provided,"
+        " attempts to find it among environment variables and"
+        " ~/.aws/config file will be made",
+    )
+    parser.add_argument(
+        "--default-region",
+        default="us-east-1",
+        help="Default region for Amamzon services. If no region is provided,"
+        " attempts to find it among environment variables and"
+        " ~/.aws/config file will be made",
```
nathanegillett, line 141: This is no longer accurate since the default is now set.

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+        r.raise_for_status()
+        total_size = int(r.headers.get("content-length", 0))
+        t = tqdm(desc=url, total=total_size, unit="iB", unit_scale=True)
+
+        # the file is deleted as soon as it is closed.
+        temp_file = tempfile.NamedTemporaryFile(delete=True)
+        for chunk in r.iter_content(chunk_size=8192):
+            if chunk:  # filter out keep-alive new chunks
+                t.update(len(chunk))
+                temp_file.write(chunk)
```
nathanegillett, line 237: Looks like a good opportunity to calculate the downloaded file's checksum.

support/reftest/data.yml
```diff
@@ -0,0 +1,22 @@
+prod-cdn-url:
+  https://cdn.redhat.com
+
+test_data:
+- path: /content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2
+  sha256: 647b38c2222caecd0859f97a0ef0ef78f249b8d96af437552d6b2950056f261e
```
nathanegillett, line 6: Are these checksums guaranteed constant? There's no possibility that a the above path will be updated at any point?

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+
+def prepare(db, s3, config, opt):
+    for item in config.test_data:
+        url = config.prod_cdn_url + item["path"]
+        # download test data to local with a name "NamedTemporaryFile"
+        temp_file = download_to_local(url, opt.key, opt.cert, opt.cacert)
+        # push test data to s3 and dynamodb
+        s3.upload_from_localfile(
+            temp_file.name, item["sha256"], item["content-type"]
+        )
```
nathanegillett, line 250: Related to my comment on data.yaml, we may want to calculate the temp_file's checksum to use here instead of betting that it'll be consistent.

support/reftest/reftest
```diff
@@ -0,0 +1,296 @@
...
+
+
+def clean(db, s3, config):
+    for item in config.test_data:
+        key = db.search_object_key_from_web_uri(item["path"])
+        if not key:
+            print("{} not found in table {}".format(item["path"], db.table))
+            continue
+        db.delete_item_from_web_uri(item["path"])
+        s3.delete_from_key(item["sha256"])
```
nathanegillett, line 263: Again, it looks like we'll always delete the "right" file from the s3 bucket but will delete a pseudo-random item from the db table. **_If_** the checksum given to delete_from_key() is accurate, we should be able to also give it to delete_item_from_web_uri() so that we know we're deleting the right table item.

dichen16: @nathanegillett you are talking about a `delete_item_from_checksum()` or `delete_item_from_object_key`, right?

nathanegillett: Yeah. Deleting items by object_key is 100% accurate, while deleting by web_uri could be a gamble.

nathanegillett: See my other comments as well. Hopefully it all makes sense together.

support/reftest/data.yml
rohanpm: Most of them should never change.  'listing' files are the most likely exception here, I think, but there is no specific reason those here ought to be updated either (it's just that they "could" have formatting changes due to a new version of tools, or something like that).

The way I see it, the test data definition needs to balance between:

1. Don't define any checksums: then we don't really know what data we're testing, which means tests can't make any assumptions about it, so you can't directly encode an expected value of Digest headers for instance.
2. Define the checksums: then we know exactly our testdata, so our tests can be coded directly against that which allows them to be simple.  But, in exchange, if the source data changes, this config file needs maintenance.

I'm pretty comfortable with using (2) for everything we currently have in here, but I did think maybe we would have to end up with the script supporting both cases, with some data using (1) and some using (2).  e.g. a repomd.xml file seems inappropriate for (2) since any republish of the repo will change it.

support/reftest/reftest
rohanpm: +1

This is pretty important to detect if reftest.yml has become outdated.

support/reftest/reftest
rohanpm: I think we should delete DynamoDB items using *both* web uri and checksum to find the item for deletion.

I'm not sure we should be deleting stuff from S3 at all.  There is a many-to-one relationship between web URIs and S3 objects, a given object could be referenced by loads of URIs including some not mentioned in reftest.yml.  Particularly for simple file formats like 'listing' files.

Example: https://cdn.redhat.com/content/dist/rhel/server/5/5.7/listing content is

```
i386
ia64
x86_64
```

If you deleted S3 object for this, you would be deleting the object for *every* listing file with that content, of which there will be many, not only the one defined in the testdata.  Similarly if you deleted DynamoDB entries while looking only at the object key and not the URI.

I'm still not really clear on what the desired behavior of 'clean' is here, but if you make it delete S3 objects then it's only safe if the environment is more or less dedicated to hold only this testdata and nothing else.  I didn't expect it to be used that way.  I think it'd be better not to delete the objects, so this can be used on an environment which is also used for other stuff.

dichen16:
@rohanpm my purpose for this "clean" operation is to "undo the prepare operation", I did not think about the many-to-one relationship between web URIs and S3 objects, 

I will change it to "undo", what do you think?

I don't want to bring dangerous operation in the "undo" operation in case it delete S3 object wrongly.
I am think about a "reference-count"  before deleting the S3 object. Specifically, deleting the S3 object only if there is one web_uri reference this object_key.

But I also don't want to make the script too complicated or make the test data different(add extra attribute), need some suggestion.

nathanegillett:
@dichen16, when you upload an object to s3, you can set metadata for it (in the upload_file method). I believe what @rohanpm's suggesting is to add a unique key-value to the uploaded object's metadata. That way you know with absolute certainty which objects are safe to remove from the bucket. **Note:** ensure that uploads aren't overriding existing objects

I'm not sure that counting the number of references to the s3 object in the table is all that reliable, as there's a (slight) chance that number could change during the operation (e.g., someone adding a table item at the same time). Scanning a table to determine what should be deleted from a bucket _seems_ like it be much less efficient than just checking an s3 object's metadata before deletion.

rohanpm:
I meant to set attributes on the DB items and not the S3 objects.  I think we just shouldn't delete the objects from S3 at all.  And I agree with Nathan that reference counting is probably not feasible.

dichen16:
@nathanegillett, @rohanpm thank you,
> Related to my comment on data.yaml, we may want to calculate the temp_file's checksum to use here instead of betting that it'll be consistent.
> This is pretty important to detect if reftest.yml has become outdated.

For outdated test data check, I will create another PR for it

support/reftest/reftest
```diff
@@ -0,0 +1,282 @@
...
+        s3.upload_from_localfile(
+            temp_file.name, item["sha256"], item["content-type"]
+        )
+        db.put_item(item["path"], item["sha256"], opt.release_date)
+        # delete the NamedTemporaryFile
+        temp_file.close()
+
+
+def undo(db, config):
+    for item in config.test_data:
```
rohanpm, line 244: I still think the undo process here is not right.  This isn't limited to undoing whatever the command previously did, it's simply going to delete entries matching the paths and checksums of the testdata - whether or not they were put in place by this command.  And given that we picked real data at real paths, it's not at all unlikely that these will exist without having been put there by this script.  Listing files are the simplest example, e.g. if someone runs rhsm-update-listings against stage, it's likely to put /content/dist/rhel/server/5/5.7/listing in place with content exactly matching the reference test data.  This command shouldn't delete that entry just because it happens to match.

Can you please somehow make undo really just undo what the command did (e.g. by storing an extra piece of metadata flagging the entry as created by this command, as I mentioned earlier), or otherwise just remove the undo command?

support/reftest/reftest
nathanegillett: This still isn't correct.

support/reftest/reftest
nathanegillett: Consensus was that we should use temp_file's checksum.

rohanpm: Just to clarify as there still seems to be a bit of confusion here, what I think this needs to do is:

- calculate the temp file's checksum
- verify that the file's checksum matches item["sha256"] - if not, it's a fatal error

Obviously then when you actually do the insert, it doesn't matter whether you use item["sha256"] or the temp file's checksum because you made sure that they are the same value.

dichen16:
**Positive Test**
```
(cdn_lambda_env) [dichen@dilaptop reftest]$ ./reftest prepare --table dichen-test --bucket dichen-test --cert /home/dichen/certs/rcm-debug/rcm-debug.crt --key /home/dichen/certs/rcm-debug/rcm-debug.key
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2: 100%|███████████████| 12.0M/12.0M [00:04<00:00, 2.60MiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz: 100%|███████████████| 25.9M/25.9M [00:12<00:00, 1.99MiB/s]
https://cdn.redhat.com/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm: 100%|█████████████████████████████████████████████| 196k/196k [00:01<00:00, 177kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/5.7/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████| 17.0/17.0 [00:00<00:00, 11.1kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/5.6/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████| 17.0/17.0 [00:00<00:00, 9.32kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████████| 30.0/30.0 [00:00<00:00, 18.9kiB/s]
prepare operation has finished successfully!
```

**Negative Test**
```
(cdn_lambda_env) [dichen@dilaptop reftest]$ ./reftest prepare --table dichen-test --bucket dichen-test --cert /home/dichen/certs/rcm-debug/rcm-debug.crt --key /home/dichen/certs/rcm-debug/rcm-debug.key
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2: 100%|███████████████| 12.0M/12.0M [00:04<00:00, 2.64MiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz: 100%|███████████████| 25.9M/25.9M [00:10<00:00, 2.51MiB/s]
https://cdn.redhat.com/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm: 100%|█████████████████████████████████████████████| 196k/196k [00:01<00:00, 182kiB/s]
/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm verify checksum failed
Fatal error: prepare operation is terminated
```

support/reftest/reftest
```diff
@@ -0,0 +1,249 @@
...
+
+
+class DBHandler:
+    def __init__(self, table, session):
+        self.dynamodb = session.client("dynamodb")
+        self.table = table
+
+    def put_item(self, web_uri, object_key, from_date=None, metadata={}):
+        if not from_date:
+            from_date = datetime.utcnow().isoformat(timespec="seconds")
```
rohanpm, line 29: Minor: rather than calculating from_date internally here, it'd probably be better to calculate it only once for the runtime of the process.  That way you can easily find the entire set of entries added by a single run of the process as they'll all have the same timestamp.

support/reftest/reftest
```diff
@@ -0,0 +1,249 @@
...
+                temp_file.flush()
+        t.close()
+        return temp_file
+
+
+# verify cdn data checksum mathch test data checksum
+def verify_sha256_checksum(temp_file_path, test_data_checksum):
+    if os.path.isfile(temp_file_path):
+        sha256 = hashlib.sha256()
+        with open(temp_file_path, "rb") as binary:
```
rohanpm, line 186: It's not really a big deal, but this is wasteful because you already read the entire file (while downloading it) and now you're reading the whole thing again.

Nathan already pointed out above in the comment "Looks like a good opportunity [...]" exactly where we could calculate checksum.  Doing it there would avoid having to re-read the file we've just written unnecessarily.

That being said, I doubt it has any notable impact on performance.

support/reftest/reftest
```diff
@@ -0,0 +1,249 @@
...
+
+def prepare(db, s3, config, opt):
+    for item in config.test_data:
+        url = config.prod_cdn_url + item["path"]
+        # download test data to local with a name "NamedTemporaryFile"
+        temp_file = download_to_local(url, opt.key, opt.cert, opt.cacert)
+
+        # checksum verify
+        if not verify_sha256_checksum(temp_file.name, item["sha256"]):
+            print("{} verify checksum failed".format(item["path"]))
```
rohanpm, line 209: I think you should at least log the actual checksum here.  If you don't then it seems like the problem can't be investigated further afterward, and you don't have the checksum you need in order to update the reference test data (because the temp file was already deleted by the time the script exits...)

support/reftest/reftest
```diff
@@ -0,0 +1,249 @@
...
+    s3_client = S3Handler(bucket=opt.bucket, session=session)
+
+    res = False
+    if opt.command == "prepare":
+        res = prepare(db_client, s3_client, config, opt)
+
+    if res:
+        print("{} operation has finished successfully!".format(opt.command))
+    else:
+        print("Fatal error: {} operation is terminated".format(opt.command))
```
rohanpm, line 245: Can you please make sure it exits with non-zero exit code in this case?

Otherwise if someone tries to do something like `support/reftest/reftest && run_tests`, it'll go ahead and start run_tests even if the reftest command failed...

dichen16:
**Exit Code Check**
```
(cdn_lambda_env) [dichen@dilaptop reftest]$ ./reftest prepare --table dichen-test --bucket dichen-test --cert /home/dichen/certs/rcm-debug/rcm-debug.crt --key /home/dichen/certs/rcm-debug/rcm-debug.key
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2: 100%|███████████████| 12.0M/12.0M [00:09<00:00, 1.32MiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz: 100%|████████████████| 25.9M/25.9M [00:28<00:00, 908kiB/s]
https://cdn.redhat.com/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm: 100%|█████████████████████████████████████████████| 196k/196k [00:01<00:00, 125kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/5.7/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████| 17.0/17.0 [00:00<00:00, 10.7kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/5.6/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████| 17.0/17.0 [00:00<00:00, 11.1kiB/s]
https://cdn.redhat.com/content/dist/rhel/server/5/listing: 100%|██████████████████████████████████████████████████████████████████████████████████████████| 30.0/30.0 [00:00<00:00, 19.6kiB/s]
prepare operation has finished successfully!
(cdn_lambda_env) [dichen@dilaptop reftest]$ echo $?
0
(cdn_lambda_env) [dichen@dilaptop reftest]$ vim data.yml 
(cdn_lambda_env) [dichen@dilaptop reftest]$ ./reftest prepare --table dichen-test --bucket dichen-test --cert /home/dichen/certs/rcm-debug/rcm-debug.crt --key /home/dichen/certs/rcm-debug/rcm-debug.key
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/fd23895d43f54a50bbd0509809dd5f45298bfd6b-other.sqlite.bz2: 100%|███████████████| 12.0M/12.0M [00:10<00:00, 1.15MiB/s]
https://cdn.redhat.com/content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz: 100%|███████████████| 25.9M/25.9M [00:17<00:00, 1.48MiB/s]
/content/dist/rhel/server/7/7.4/x86_64/os/repodata/cb753af26534673064bd593500d747d7288d75b2-filelists.xml.gz verify checksum failed, cdn_data_checksum is 49159602b5bb8d9d34af384a62bd411d12449338501a4d08ef90c369b7facb98, but test_data_checksum is 49159602b5bb8d9d34af384a62bd411d12449338501a4d08ef90c369b7facb98_
Fatal error: prepare operation is terminated
(cdn_lambda_env) [dichen@dilaptop reftest]$ echo $?
1

```

PR #62, opened 2020-05-18
Consolidate CodeBuild configuration in pipeline template

nathanegillett:
Previously, separate files were kept for the main CodeBuild buildspec
and the verify-source CodeBuild project. Since we can write buildspecs
in-line, there doesn't seem to be a benefit to this separation.

So, this commit moves the verify-source project and buildspec for the
build action into the pipeline template, reducing the number of files
maintained within the configuration directory.


nathanegillett:
@rohanpm, @crungehottman, @dichen16, do you guys have any concerns or questions about this?

PR #63, opened 2020-05-19
Parameterize CloudFormation templates [RHELDST-1035] 

crungehottman:


configuration/exodus-pipeline.yaml
```diff
@@ -273,14 +284,14 @@ Resources:
...
+    Condition: IsExodusProd
     Properties:
-      S3BucketName: exodus-pipeline-artifacts
+      S3BucketName: !Sub cloudtrail-logs
       EventSelectors:
         - DataResources:
           - Type: AWS::S3::Object
             Values:
-              - arn:aws:s3:::exodus-pipeline-artifacts/build-stage
+              - "arn:aws:s3:::"
```
nathanegillett, line 294: What's the intention here?

Remember that prod pipeline is triggered by changes to this object. So, I doubt the trail can be shared by multiple pipeline sets.

configuration/exodus-pipeline.yaml
```diff
@@ -46,6 +50,11 @@ Conditions:
     !Equals [!Ref env, stage]
   IsProd:
     !Equals [!Ref env, prod]
+  IsExodus:
+    !Equals [!Ref project, exodus]
+  IsExodusProd: !And
+    - !Condition IsExodus
+    - !Condition IsProd
```
nathanegillett, line 57: super minor: I'd prefer all the conditions have a consistent format

crungehottman: By consistent, were you referring to not referencing another Condition, or consolidating the condition definition to a single line? 
```
  IsExodusProd: 
    !And [!Equals [!Ref project, exodus], !Equals [!Ref env, prod]]
```
vs 
```
  IsExodusProd: 
    !And [!Condition IsExodus, !Condition IsProd]
```
If the former, is there a reason why we shouldn't reference another condition?

nathanegillett: I was referring to the latter, single vs multi-line, just because all other conditions in the repo are formatted that way. Your second example looks great.

configuration/exodus-pipeline.yaml
```diff
@@ -273,15 +283,16 @@ Resources:
...
         - DataResources:
           - Type: AWS::S3::Object
             Values:
-              - arn:aws:s3:::exodus-pipeline-artifacts/build-stage
+              - "arn:aws:s3:::"
           ReadWriteType: WriteOnly
+          IncludeManagementEvents: false
       IncludeGlobalServiceEvents: true
       IsLogging: true
       IsMultiRegionTrail: true
```
nathanegillett, line 298: Monitoring all s3 actions across the account seems like it would get very expensive very fast (log storage in s3).
I don't think it's really necessary either, as our only use for this right now is for pipeline activities. CloudTrail doesn't need to know about every change in the CDN buckets, for instance -- no action needs to be taken for those events.

According to lines 275-277, every "project" is using *-pipeline-artifacts/build-stage as the object for triggering its prod pipeline, so I'm not sure why the EventSelectors needs this change anyway. Can't it just be `arn:aws:s3:::${project}-pipeline-artifacts/build-stage`?

configuration/exodus-pipeline.yaml
```diff
@@ -305,7 +316,39 @@ Resources:
...
+      BucketName: cloudtrail-logs
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
+      LifecycleConfiguration:
+        Rules:
+          - Status: "Enabled"
+            ExpirationInDays: 1
```
nathanegillett, line 344: The documentation I found was unclear to me; am I correct in assuming this will permenently delete all logs after 24 hours?

configuration/exodus-pipeline.yaml
```diff
@@ -273,25 +283,26 @@ Resources:
...
+    Condition: IsExodusProd
     Properties:
-      S3BucketName: exodus-pipeline-artifacts
+      S3BucketName: cloudtrail-logs
       EventSelectors:
         - DataResources:
           - Type: AWS::S3::Object
             Values:
-              - arn:aws:s3:::exodus-pipeline-artifacts/build-stage
+              - "arn:aws:s3:::"
```
nathanegillett, line 293: I still can't come up with a good reason to monitor all s3 activity.

configuration/exodus-pipeline.yaml
```diff
@@ -273,25 +283,26 @@ Resources:
...
           ReadWriteType: WriteOnly
+          IncludeManagementEvents: false
       IncludeGlobalServiceEvents: true
       IsLogging: true
-      IsMultiRegionTrail: true
```
nathanegillett, line 287: This got me curious and, after looking into it, I don't think we need IsLogging or IncludGlobalServiceEvents either.

configuration/exodus-pipeline.yaml
crungehottman: > Monitoring all s3 actions across the account seems like it would get very expensive very fast (log storage in s3).
> I don't think it's really necessary either, as our only use for this right now is for pipeline activities. CloudTrail doesn't need to know about every change in the CDN buckets, for instance -- no action needs to be taken for those events.

If we decide to use a separate AWS account for the exodus dev env pipeline (as briefly discussed/suggested in one of our sync meetings a while ago), we could also use this account for our developer testing pipelines. I don't suppose we would have too many s3 bucket writes in our development account. If we have 5 million s3 object writes per month, storage costs alone would be around $0.50/mo. API calls to the s3 buckets for 5 million items would be $50/mo (deletes are free). CloudTrail pricing for 5 million s3 object writes would also be $50/mo. I don't know how many s3 write events we would do per month on a production account, but I would guess less than 100,000 on our development accounts per month. So, CloudTrail costs on our development env would would be quite low. I have also added a lifecycle rule to the cloudtrail logs bucket that will delete the logs after one day to further reduce costs. In such a case, we could use one cloudtrail for the developement aws account that monitors all writes to buckets, and one for the production account (that contains the stage and prod envs) that monitors only the exodus-pipeline-artifacts bucket. 

> According to lines 275-277, every "project" is using *-pipeline-artifacts/build-stage as the object for triggering its prod pipeline, so I'm not sure why the EventSelectors needs this change anyway. Can't it just be arn:aws:s3:::${project}-pipeline-artifacts/build-stage?

No, because using a variable name in the CloudTrail bucket definition will either create a new CloudTrail trail (in the event that a name is not specified for the resource) for the {project}-pipeline, or, if we assign the CloudTrail resource a name, and try to update it with arn:aws:s3:::${project}-pipeline-artifacts/build-stage for each project, CloudFormation will complain that the resource already exists in another stack, and will refuse to update the resource. There are other options. We could maintain a list of developers' buckets in our CloudFormation templates, or share one or two pipeline-artifacts buckets. We could also disable the automatic pipeline run on developers' prod pipelines. I've outlined these options in my comment in RHELDST-1035.

> Is there some other use case in which broader monitoring is necessary?

I do not have a specific idea in mind, but it might be useful to monitor API calls on our cdn buckets. It seems like it could be useful for security and/or auditing purposes.

configuration/exodus-pipeline.yaml
crungehottman: Correct

crungehottman:
I still have a couple of minor things to fix, but in the meantime I request @rohanpm's feedback on the general solution.

configuration/exodus-pipeline.yaml
crungehottman: Hmm, I am not sure about IsLogging. I thought IsLogging defines whether the trail is enabled or disabled (see each trail's "status" field [here](https://console.aws.amazon.com/cloudtrail/home?region=us-east-1#/dashboard), or the on/off switch within a Trail). What is your understanding of this bool? Either way, +1 to removing IncludeGlobalServiceEvents.

configuration/exodus-pipeline.yaml
```diff
@@ -46,6 +50,10 @@ Conditions:
     !Equals [!Ref env, stage]
   IsProd:
     !Equals [!Ref env, prod]
+  IsExodus:
```
rohanpm, line 53: I'm not really a fan of this part, since part of the reason to provision this with a different project in the first place would be testing the pipeline, but forcibly disabling chunks of it whenever project isn't named "exodus" makes it impossible to test those parts.

Is it possible to have an explicit parameter instead, named after what it does?  e.g. useCloudtrails with default of false?

configuration/exodus-pipeline.yaml
nathanegillett: Ah, I hadn't considered a change to the event arn creating a new trail. So, yeah, I suppose if we want to use this for all pipelines we either need to monitor all buckets or force all the pipelines to use a single bucket, which probably wouldn't work out well.

configuration/exodus-pipeline.yaml
nathanegillett: Oh, I think you're right about IsLogging. I misread the doc for it.

configuration/exodus-pipeline.yaml
nathanegillett: The problem is that, once created as part of a CloudFormation stack, the CloudTrail trail will be locked to that stack.
So, if a dev creates a stack to test the pipeline before the real pipelines are created, the dev would have to delete their stack, or at least break it by deleting the trail, before the real pipelines can go up.

With having `IsExodusProd`, we ensure that the trail is tied to the real environment, and devs can then use the trail in their own pipelines afterward. The only thing you'd be unable to test post-creation is the creation iteself.

crungehottman: > The only thing you'd be unable to test post-creation is the creation itself.

+1.  In theory, if our account uses a single trail to monitor all s3 write API calls, it should be a very rare occurrence for us to ever modify our existing trail, thus reducing the need to test the creation of the CloudTrail resource.

Though, if strongly preferred, is possible for us to maintain separate trails. I can prepend the ${project} var to each resource so they can be created in separate stacks. We will just have to be aware that we only have room for two development CloudTrail trails, and as such, only two developers can test their stage -> prod transitions in parallel. Thus, we will need to strictly enforce that each developer only create their CloudTrail resources immediately before testing, and delete their CloudTrail resources immediately after testing. This could be accomplished by deploying the developer's prod pipeline with the "useCloudtrails=true" param before testing, and then without the "useCloudtrails" param after testing. The only real negative of this solution is a limited number of resources, and the strict requirement of enforcing clean-up. Maintaining separate trails would also enable us to reduce the data we collect (reducing the monitoring to only one {project}-pipeline-artifacts bucket per trail), as Nathan suggested in an earlier comment. 

Note that the reduction in data collection is only useful if we do not want to collect data for s3 write API calls in our production exodus AWS account. If we want to collect data for all s3 write API calls for security/auditing reasons, I think we should stick with the single trail and accept that the creation of a shared, account-wide resource is not something a developer will not need to test.

nathanegillett: I suppose it makes some sense that one may want to test the prod pipeline by itself, without promotion from stage. So, I'm okay with having a CloudTrail parameter if we're set on multiple trails.

However, I don't like the idea of steering devs into the AWS console/cli and delete individual resources -- seems like it could get messy. We should probably try to limit interaction with resources to CloudFormation create/update, delete. I.e., if one enables CloudTrail with their pipeline stack, the trail should be deleted with the their pipeline stack. I'm not sure if something like that's enforceable, though.

nathanegillett: If it matters, I'm in favor of the single trail with a log expiration of 1 day. We may have other uses for CloudTrail in the future.

rohanpm: When writing my earlier comments, I didn't really understand how the cloudtrail was connected to our other resources.  Now taking the time to try to understand it, based on the EventSelectors it seems that one trail is going to monitor events for resources provisioned from *all* values of ${project} and ${env} so there really is no point to permanently maintain more than one.

I might not have objected to IsExodusProd if I understood that at the time.  Nevertheless, I like the way it is now which does seem to support temporarily provisioning a new cloudtrail if needed.  A reason why it might be needed is if we want to try changing the settings (e.g. the EventSelector) on the trail.  It would be better to create a separate one and try out the proposed config there rather than immediately deploying the changes to the single production trail.

nathanegillett:
So, we're going with the multiple trail approach? I'm a little worried about dev/test trails blocking other uses for CloudTrails as a result but I suppose we can change it down the line if it becomes a problem.

> Maintaining separate trails would also enable us to reduce the data we collect (reducing the monitoring to only one {project}-pipeline-artifacts bucket per trail)

If we're doing individual trails, can we revert the EventSelector value as you've described?

crungehottman:
> So, we're going with the multiple trail approach? I'm a little worried about dev/test trails blocking other uses for CloudTrail as a result but I suppose we can change it down the line if it becomes a problem.
> 
> > Maintaining separate trails would also enable us to reduce the data we collect (reducing the monitoring to only one {project}-pipeline-artifacts bucket per trail)
> 
> If we're doing individual trails, can we revert the EventSelector value as you've described?
> We can keep the expiration too, I think.

@nathanegillett , I think we're going to go with the single trail, but we will enable devs to create a separate, temporary trail only when completely necessary. I've prepended "${project}-" on all of the CloudTrail resources just to enable--not to enforce or suggest--a developer's temporary trail creation. The creation of a separate trail is disabled by default. There should be no need for a dev to enable the creation of a separate CloudTrail resource, as the dev's CloudWatch event will automatically use the exodus pipeline's trail, which acts on all buckets. I could write up some "dev testing guide" documentation to help reduce the number of instances in which a developer creates a CloudTrail and to instruct them on how to clean up the resources after testing.

nathanegillett:
Oh, I see now, @crungehottman. Thanks.

PR #65, opened 2020-05-22
Resolve resource dependencies in the CloudFormation templates [RHELDST-1035]

crungehottman:


nathanegillett:
Should probably include the reason for this change in the commit message, as, on the surface, duplicating the roles seems bad.

crungehottman:
Updated my commit message.

PR #69, opened 2020-05-28
Use real branch name in verify-source [RHELDST-1288]

nathanegillett:
This commit changes verify-source's REPO_BRANCH variable to the branch
name from the source stage rather than the template parameter value.

This resolves an issue wherein no branch is specified by the user,
preferring to instead use the template's default values, and an attempt
to checkout "None" as a branch is made.

### DIFF
--- a/configuration/exodus-pipeline.yaml
+++ b/configuration/exodus-pipeline.yaml
@@ -186,7 +186,7 @@ Resources:
                       !Sub '[
                         {"name": "REPO_OWNER", "value": "${repoOwner}", "type": "PLAINTEXT"},
                         {"name": "REPO_NAME", "value": "${repoName}", "type": "PLAINTEXT"},
-                        {"name": "REPO_BRANCH", "value": "${repoBranch}", "type": "PLAINTEXT"},
+                        {"name": "REPO_BRANCH", "value": "#{SourceVars.BranchName}", "type": "PLAINTEXT"},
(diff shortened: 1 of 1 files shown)


dichen16:
```
aws cloudformation deploy --template-file configuration/exodus-pipeline.yaml --stack-name $PROJECT-pipeline-$env --capabilities CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND  --parameter-overrides  env=$env  repoOwner=dichen16  repoName=exodus-lambda  repoBranch=master     region=us-east-1     email=project-exodus@redhat.com     project=$PROJECT     useCloudTrail=false     oai='EOBHR9XSAFIWV'  githubToken=$mytoken

Waiting for changeset to be created..
Waiting for stack create/update to complete
Successfully created/updated stack - dichenx-pipeline-dev
```
It works but I think the option `repoBranch=master` is not used.

nathanegillett:
Why don't you think the repoBranch value is used, @dichen16?

dichen16:
@nathanegillett  cuz the ${repoBranch} is removed, so I think the option `repoName=exodus-lambda` I put in the code block is useless. I should not use it in the command.

nathanegillett:
@dichen16, `${repoBranch}` was only replaced in verify-source's buildspec. It's still used by the webhook and GitHub source actions.

PR #72, opened 2020-06-02
Add integration tests to pipeline [RHELDST-719]

crungehottman:
This commit introduces an additional "Test" stage to the
stage and prod lambda pipelines.

The "Test" stage contains an "IntegrationTests" action,
which runs the integration tests against the environment's
CloudFront distribution via a CodeBuild project which clones
the git repo, determines the distribution's domain name, and
runs the integration tests.

If the integration tests pass, the BuildArtifact of the successful
run is preserved in the pipeline-artifacts bucket as build-${env}.

If the integration tests fail, SNS notification is sent and a
manual rollback is triggered. The manual rollback is triggered by
a CloudWatch Event Rule that uses a CodeBuild project as a target.
The CodeBuild project deploys the CloudFormation template from the
BuildArtifact of the last successful run.


configuration/exodus-pipeline.yaml
```diff
@@ -58,6 +58,8 @@ Conditions:
     !Equals [!Ref env, prod]
   NotProd:
     !Not [!Equals [!Ref env, prod]]
+  NotDev:
+    !Not [!Equals [!Ref env, dev]]
```
nathanegillett, line 62: Super minor: I wish that these Not{env} conditions followed the same ordering as the Is{env} conditions

configuration/exodus-pipeline.yaml
```diff
@@ -230,6 +232,27 @@ Resources:
                 TemplatePath: BuildArtifact::exodus-lambda-pkg.yaml
               InputArtifacts:
                 - Name: BuildArtifact
+        - !If
+          - NotDev
+          - Name: Test
```
nathanegillett, line 237: I think we should run integration tests before deploying to dev as well.

configuration/exodus-pipeline.yaml
```diff
@@ -260,9 +283,26 @@ Resources:
                   - Name: BuildArtifact
                 RunOrder: 2
           - !Ref AWS::NoValue
+        - !If
+          - IsProd
+          - Name: Upload
```
nathanegillett, line 288: I think we should consolidate uploading here (switch to !If NotDev and delete Upload from Promote stage).

configuration/exodus-pipeline.yaml
```diff
@@ -260,9 +283,26 @@ Resources:
...
+                Configuration:
+                  BucketName: !Sub ${project}-pipeline-artifacts
+                  ObjectKey: !Sub build-${env}
+                  Extract: false
+                InputArtifacts:
+                  - Name: BuildArtifact
+          - !Ref AWS::NoValue
     DependsOn:
       - CodePipelineRole
-      - CodePipelineBucket
```
nathanegillett, line 265: Did you mean to remove this? Should definitely remain.

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +477,113 @@ Resources:
...
+              commands:
+                - pip install --upgrade pip
+                - pip install tox
+            build:
+              commands:
+                - git clone https://github.com/$REPO_OWNER/$REPO_NAME.git -b $REPO_BRANCH
+                - cd ./$REPO_NAME
+                - virtualenv integration_tests
+                - source integration_tests/bin/activate
+                - tox -e py37 ./tests/integration -- --cdn-test-url https://$(aws cloudformation describe-stacks --stack-name $STACK_NAME | jq -r '.Stacks[0].Outputs[] | select(.OutputKey=="Distribution").OutputValue')
```
nathanegillett, line 507: Can we break the tox command down a little so it's a little easier to read? Like, maybe a variable to hold the URL?

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +477,113 @@ Resources:
       TimeoutInMinutes: 10
     DependsOn: CodeBuildRole
 
+  IntegrationTestsCodeBuildProject:
+    Type: AWS::CodeBuild::Project
+    Condition: IsDev
```
nathanegillett, line 482: Currently, this won't ever run, as test stage is only created for stage, prod.

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +477,113 @@ Resources:
...
+                - source integration_tests/bin/activate
+                - tox -e py37 ./tests/integration -- --cdn-test-url https://$(aws cloudformation describe-stacks --stack-name $STACK_NAME | jq -r '.Stacks[0].Outputs[] | select(.OutputKey=="Distribution").OutputValue')
+      Artifacts:
+        Type: CODEPIPELINE
+      TimeoutInMinutes: 15
+
+  RollbackLambdasCodeBuild:
+    Type: AWS::CodeBuild::Project
+    # Create CodeBuild project for running integration tests against the dev pipeline
+    Condition: NotDev
```
nathanegillett, line 515: So, as I suggested earlier, we should probably run integration tests for dev and therefore enable rollback for it too.

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +477,113 @@ Resources:
...
+              "FAILED"
+            ],
+            "stage": [
+              "Test"
+            ],
+            "action": [
+              "IntegrationTests"
+            ]
+          }
+        }'
```
nathanegillett, line 573: I don't think it's necessary to inject JSON here, YAML formatting should work fine.
> resources:
     - !Sub arn:aws:codepipeline:${region}:${AWS:AccountId}:${CodePipeline}

configuration/exodus-pipeline.yaml
```diff
@@ -532,7 +675,7 @@ Resources:
 
   CloudWatchEventRole:
     Type: AWS::IAM::Role
-    Condition: IsProd
+    Condition: NotDev
```
nathanegillett, line 678: I think this would need to go too if running integration tests and rollback in dev

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +477,113 @@ Resources:
...
+          phases:
+            install:
+              runtime-versions:
+                python: 3.7
+              commands:
+                - pip install --upgrade pip
+                - pip install tox
+            build:
+              commands:
+                - git clone https://github.com/$REPO_OWNER/$REPO_NAME.git -b $REPO_BRANCH
```
rohanpm, line 503: The branch content could have changed already after the pipeline started and before we get to here.  Should it use a specific commit instead?

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +477,113 @@ Resources:
...
+                - cd ./$REPO_NAME
+                - virtualenv integration_tests
+                - source integration_tests/bin/activate
```
rohanpm, line 506: You tested it, right?  Just asking because actually this is only going to work if every command is run in the same shell, and it's not obvious to me that it would work that way.

e.g. if the behavior in AWS is that each element in 'command' is run via "/bin/sh -c <cmd>", then this isn't going to do anything useful since it's only changing environment in the current shell, which would then immediately exit.

Actually, what does this virtualenv do?  We installed tox *outside of* the virtualenv, so it doesn't seem like activating it is necessary to get a 'tox' in path.  Were you expecting tox to install packages into this virtualenv?  tox normally creates its own virtualenvs as needed.

configuration/exodus-pipeline.yaml
rohanpm: Can we put as much as possible into tox.ini instead?

Remember updating the pipeline is more complicated than just merging a commit into the repo.  If we need to change the test command here, like adding or changing the arguments passed to pytest, it would be better if we can just do that by tweaking tox.ini and not having to redeploy the pipeline.

I think ideally it would just be "tox -e integration-tests", if feasible.

configuration/exodus-pipeline.yaml
rohanpm: +1

configuration/exodus-pipeline.yaml
crungehottman: Yes. I filed https://projects.engineering.redhat.com/browse/RHELDST-1363 to discuss this issue.

configuration/exodus-pipeline.yaml
crungehottman: It will. This creates the CodeBuild project, which is independent from the CodePipeline, so it doesn't matter which env this resource is created in, so long as it's created in some env before it's referenced (i.e., <= stage). When this template is deployed against dev, the CodeBuild project is created. When the template is deployed against stage and prod, it references the CodeBuild project that was already created in the dev stage. The same method was used for the other CodeBuild project (Condition: !IsDev when in practice, this project is used in both the dev and stage pipelines because we only need one of these to be created). Ideally, we shouldn't use inaccurate "IsDev" conditions to represent or satisfy an "only create this resource once even though this resource is going to be used in multiple envs" condition. A separate template/stack is a better fit for such resources. I've addressed that here: https://projects.engineering.redhat.com/browse/RHELDST-1363 .

configuration/exodus-pipeline.yaml
crungehottman: this condition is no longer necessary as int tests will be run against all envs

configuration/exodus-pipeline.yaml
nathanegillett: Ah, right. Yeah...maybe we should have a separate stack of shared resources. This template has become much more complex than it initially was.

configuration/exodus-pipeline.yaml
nathanegillett: Ah, yes, I missed this yesterday.

The `$REPO_BRANCH` variable will be "None" in most cases so that the default master/deploy branches are used, so this won't work. Instead of targeting a branch, I think we'll want to checkout #{SourceVars.CommitId}, which is the specific commit verified and deployed for each pipeline.

Note that the SourceVars namespace is only available in dev and stage pipelines, so this CodeBuild project will need to be created in one of those stacks.

configuration/exodus-pipeline.yaml
nathanegillett: From my experience, the shell seems to be specific to the phase. E.g., don't expect commands from a "pre-build" phase to carry through to a "build" phase. So, unless Amazon changes how this works down the line, this buildspec will work.

However...
> tox normally creates its own virtualenvs as needed

This. The additional env doesn't seem necessary.

configuration/exodus-pipeline.yaml
crungehottman: I would rather include the git repo as a secondary source in the prod pipeline.

configuration/exodus-pipeline.yaml
crungehottman: +1 to a virtualenv being unnecessary.

configuration/exodus-pipeline.yaml
```diff
@@ -120,39 +120,35 @@ Resources:
...
-          - !Ref AWS::NoValue
-        # Create S3 source for prod pipeline
-        - !If
-          - IsProd
-          - Name: Source
-            Actions:
-              - Name: Source
+        - Name: Source
+          Actions:
+            - Name: Source
```
nathanegillett, line 125: Doesn't this mean that prod pipeline will now be triggered on pushes to the deploy branch as well as updates to the template by the stage pipeline?

At least the stage pipeline is still stopped if the commit isn't verified. With prod, any time anything is pushed to the deploy branch, it'll go through an unnecessary redeployment. Prod will either run once if the push was illegitimate or twice if it was (one for the push [ineffective], one for the template update).

I don't know the full financial impact of this but it's money wasted. Anything we can do to avoid that, if so?

configuration/exodus-pipeline.yaml
```diff
@@ -438,10 +448,94 @@ Resources:
       TimeoutInMinutes: 10
     DependsOn: CodeBuildRole
 
+  IntegrationTestsCodeBuildProject:
+    Type: AWS::CodeBuild::Project
```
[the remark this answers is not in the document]
nathanegillett: I suppose reworking "CodeBuildProject" to make its function more clear is also an option...
"BuildProject", "BuildCodeBuildProject", "SourceBuildProject"? ¯\_(ツ)_/¯

configuration/exodus-pipeline.yaml
crungehottman: Good point, but the git webhook is only configured for dev and stage envs, so I don't think this should be a problem. My [prod pipeline history](https://console.aws.amazon.com/codesuite/codepipeline/pipelines/caleigh-pipeline-prod/executions?region=us-east-1) only contains executions from the trigger `CloudWatchEvent - arn:aws:events:us-east-1:333296168829:rule/caleigh-pipeline-prod-CodePipelineEventRule-2WXYIRNUXELZ`; as opposed to my [stage pipeline history](https://console.aws.amazon.com/codesuite/codepipeline/pipelines/caleigh-pipeline-stage/executions?region=us-east-1), which only contains executions from the trigger `Webhook - arn:aws:codepipeline:us-east-1:333296168829:webhook:caleigh-pipeline-stage`.

nathanegillett: Glad to be wrong. Thanks!

configuration/exodus-pipeline.yaml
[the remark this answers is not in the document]
crungehottman: +1. I agree that we should rework "CodeBuildProject" now that we have multiple CodeBuild projects. If we rename the "CodeBuildProject" logical id, we should probably also rename the CodeBuild project itself, as it is named "{project}-codebuild".

Just something to note, apparently renaming resources isn't exactly straightforward with CloudFormation. It either requires a little dance, or deletion of the resources. 

If you don't think we will build other source code through aws, I'm fine with "SourceBuildProject" (${project}-source-build). If we think we might build some other source code in the future through aws, we may want to be even more descriptive, like: "PackageLambdasProject" (${project}-package-lambdas}), "LambdaBuildProject"/"BuildLambdasProject" (${project-lambda(s)-build}, ${project}-build-lambdas}, or "BuildPackageProject/"PackageBuildProject" (${project}-package-build, ${project}-build-package}). Or, maybe even more specific, like: "Build/PackageCloudfrontLambdasProject" (${project}-build-cloudfront-lambdas/${project}-package-cloudfront-lambdas). I'm personally in favor of the most specific/self-documenting ids, but they are pretty wordy. What do you think?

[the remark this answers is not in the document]
nathanegillett: I like "LambdaBuildProject" and "${project}-lambda-build". They're concise and unique enough.

dichen16:
@crungehottman, I did not see the test data preparation, once it run the integration test without any data in dynamoDB and S3, it will fail.Or the test data preparation is triggered somewhere else?

crungehottman:
> @crungehottman, I did not see the test data preparation, once it run the integration test without any data in dynamoDB and S3, it will fail.Or the test data preparation is triggered somewhere else?

I suppose the test data preparation is triggered elsewhere. The ticket's description states:

> The integration tests are allowed to assume that all tested environments have the "reference testdata" provisioned.

tox.ini
```diff
@@ -75,3 +75,13 @@ commands=
...
 testpaths = tests
 addopts = -v
+
+[testenv:integration-tests]
+envlist = py37
+passenv = *
+deps=
+	-rrequirements.txt
+commands_pre=
+	python -m pip install -rtest-requirements.txt
```
rohanpm, line 85: Sorry to throw a wrench in the works but I think we have a security problem here.  test-requirements.txt, unlike requirements.txt, currently isn't using hash checking mode, and we're running tests in an environment where AWS credentials are available (I guess in environment variables?)

This means anyone with the ability to write to any PyPI projects in the dependency tree of our test deps can harvest our AWS credentials.

If we're going to install from test-requirements.txt in an environment which has our secrets available, can we please use hash checking mode there as well?

PR #80, opened 2020-06-16
Parameterize whitelist used for verify-source [RHELDST-1290]

nathanegillett:
Previously initial stage pipeline runs were guaranteed to fail on the
'verify' action due to missing whitelist in the pipeline's artifact
bucket. This commit implements a new resource to provision the bucket
with an initial whitelist.

If a URL and lambda function role ARN are passed to the
exodus-pipeline-resources template, a lambda function is created and
run, uploading the content at the URL to the pipeline artifact bucket.

Note: URLs should point to publicly accessible raw/plain text, e.g.,
https://raw.githubusercontent.com/.../.whitelist.gpg.

### DIFF
--- a/configuration/exodus-pipeline-resources.yaml
+++ b/configuration/exodus-pipeline-resources.yaml
@@ -18,6 +18,14 @@ Parameters:
       - false
     Default: false
     Description: Determines whether to create CloudTrail resources
+  whitelistUrl:
+    Type: String
+    Default: None
+    Description: The URL to a GPG whitelist for initializing source verification
+  lambdafunctionrole:
+    Type: String
+    Default: None
+    Description: The IAM Role ARN for Lambda Function resource
   codebuildrole:
     Type: String
     Default: None
@@ -26,6 +34,8 @@ Parameters:
 Conditions:
   CreateCloudTrail:
     !Equals [!Ref useCloudTrail, true]
+  UploadWhitelist:
+    !And [!Not [!Equals [!Ref lambdafunctionrole, None], !Not [!Equals [!Ref whitelistUrl, None]]]
 
 Resources:
   CodePipelineCloudTrail:
@@ -219,3 +229,44 @@ Resources:
       Artifacts:
         Type: NO_ARTIFACTS
       TimeoutInMinutes: 10
+
+  WhitelistUploadResource:
+    Type: Custom::WhitelistUploadFunction
+    Condition: UploadWhitelist
+    Properties:
+      ServiceToken: !GetAtt WhitelistUploadFunction.Arn
+      DestBucketName: !Sub ${project}-pipeline-artifacts
+      DestBucketRegion: !Ref region
+      WhitelistUrl: !Ref whitelistUrl
+    DependsOn:
+      - WhitelistUploadFunction
+
+  WhitelistUploadFunction:
+    Type: AWS::Lambda::Function
+    Condition: UploadWhitelist
+    Properties:
+      Handler: index.handler
+      Role: !Ref lambdafunctionrole
+      FunctionName: !Sub ${project}-whitelist-upload
+      Code:
+        ZipFile: |
+          import urllib.request
+          import cfnresponse
+          import boto3
+
+
+          def handler(event, context):
+              bucket = event["ResourceProperties"]["DestBucketName"]
+              region = event["ResourceProperties"]["DestBucketRegion"]
+              whitelist_url = event["ResourceProperties"]["WhitelistUrl"]
+
+              # Upload whitelist only when this resource is created
+              if event["RequestType"] == "Create":          
+                  s3client = boto3.client("s3", region_name=region)
+
+                  whitelist = urllib.request.urlopen(whitelist_url).read()
+                  s3client.put_object(Body=whitelist, Bucket=bucket, Key=".whitelist.gpg")
+
+              cfnresponse.send(event, context, cfnresponse.SUCCESS, {"Data": "COMPLETE"})
+      Runtime: python3.7
+    DependsOn: CodePipelineBucket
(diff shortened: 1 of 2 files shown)


configuration/exodus-pipeline-resources.yaml
```diff
@@ -219,3 +229,44 @@ Resources:
...
+              bucket = event["ResourceProperties"]["DestBucketName"]
+              region = event["ResourceProperties"]["DestBucketRegion"]
+              whitelist_url = event["ResourceProperties"]["WhitelistUrl"]
+
+              # Upload whitelist only when this resource is created
+              if event["RequestType"] == "Create":          
+                  s3client = boto3.client("s3", region_name=region)
+
+                  whitelist = urllib.request.urlopen(whitelist_url).read()
+                  s3client.put_object(Body=whitelist, Bucket=bucket, Key=".whitelist.gpg")
```
nathanegillett, line 268: @rohanpm 
So, I know you'd prefer to have the initial public keys hard-coded here but this implementation has file size limitations[1]. This means it'd be tough to fit more than a single key in this way. I think devs would have a hard time with this approach also, as they'd most likely need to hack this function for a smooth stack creation. Perhaps a compromise is to allow the user to pass in a public key string. 

[1] https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html#cfn-lambda-function-code-zipfile

CC: @crungehottman @dichen16

configuration/exodus-pipeline-resources.yaml
```diff
@@ -219,3 +229,44 @@ Resources:
...
+          import boto3
+
+
+          def handler(event, context):
+              bucket = event["ResourceProperties"]["DestBucketName"]
+              region = event["ResourceProperties"]["DestBucketRegion"]
+              whitelist_url = event["ResourceProperties"]["WhitelistUrl"]
+
+              # Upload whitelist only when this resource is created
+              if event["RequestType"] == "Create":
```
rohanpm, line 264: I'm not familiar with how the RequestType is applied.  Can this also be used to reset the whitelist on demand? e.g. imagine a commit accidentally breaks formatting of the whitelist file and locks us all out from deployment by normal means.  Can we fix it by rerunning this pointing at a correct whitelist, or would it not be possible to trigger the "Create" condition?

configuration/exodus-pipeline-resources.yaml
rohanpm: I'm not liking this, overall it doesn't feel very clean.

- the python code embedded directly here means it's not covered by all the usual quality checks on our python code, e.g. the autotests and code formatter.
- it could presumably be tested & packaged like our other lambdas, but it seems cumbersome.
- the lambda function is sitting there all the time although it's never used after initial bootstrapping
- is the lambda missing some error handling?  What happens if the URL is not fetchable - we never send any cfnresponse in that case, so does it just time out?

I have two alternative suggestions to consider:

1. Have this whitelist put in place by the ansible role rather than in this template.  It does mean that this template won't give you absolutely everything you need to have a working pipeline, which is a shame, but the code will end up so much cleaner it's probably worth it.  aws_s3 ansible module seems to provide everything we need, it would look something like:

```
aws_s3:
  bucket: "{{ exodus_artifacts_bucket }}"
  mode: put
  object: .whitelist.gpg
  # by default, only put whitelist if it's absent, but can force overwrite
  # by setting exodus_whitelist_overwrite=always
  overwrite: "{{ exodus_whitelist_overwrite|default('never') }}"
  # This would be pointed at a whitelist file checked into the playbooks repo
  src: "{{ exodus_whitelist }}"
  # (... and all the usual aws credentials etc attributes here)
```

One thing I like about this is: just as we initially deploy the pipeline by running a job in ansible tower, if the need arises to manually reset the whitelist, we just need to re-run the ansible job with a parameter asking it to reset.

2. How about, in the pipeline, if the whitelist file doesn't exist in s3 then we just don't verify?  It means the very first deployment isn't verified, which isn't ideal, but as a one-time event it seems acceptable (I think I mentioned that in discussions around the original issue).  It doesn't seem to reduce the security of regular deployments - if an attacker had permission to delete the whitelist file, then they also could have replaced it with a whitelist containing their key, so no real change there.

I don't quite like this as much as (1) since it would be nice to have verification right from the beginning, and also since the method of resetting the whitelist would seem to be "manually delete it from s3" which is worse than having an ansible job doing it.

configuration/exodus-pipeline-resources.yaml
nathanegillett: "Create" only happens once, when the resource is initially created, but "Update" is another RequestType. If "Update" was added here, the whitelist would be overwritten each time this stack is updated.

configuration/exodus-pipeline-resources.yaml
nathanegillett: Option 1 sounds best but if we're going to store a whitelist in the Ansible role repo, we may as well make that the sole copy. By this I mean removing the whitelist from release-engineering/exodus-lambda and therefore the final command of the verify-source CodeBuild project, which refreshes .whitelist.gpg in the bucket upon verification.

The whitelist in the role repo would be vaulted to prevent hijacking and serve as our single source of truth for authorizing stage & prod exodus-lambda updates. This would mean that any update to the whitelist wouldn't take effect until the playbook is run again. In other words, a newly added developer wouldn't be able to successfully trigger the stage pipeline until the playbook is re-run (and stacks updated).

Thoughts?
@rohanpm 
@crungehottman

rohanpm: Yeah, I guess we don't really need the automated updating of whitelist.gpg from this repo.  Maybe simpler is best, meaning the ansible solution.

nathanegillett: Alright. I'll close this PR and update the Jira issue.

PR #83, opened 2020-06-22
Parameterize origin access identity in exodus-storage [RHELDST-1325]

nathanegillett:
This commit removes the creation of the origin access identity, instead
requiring it as a parameter. This allows for use of exodus-storage.yaml
in exodus-lambda-playbooks.

### DIFF
--- a/configuration/exodus-storage.yaml
+++ b/configuration/exodus-storage.yaml
@@ -15,6 +15,9 @@ Parameters:
     Type: String
     Default: exodus
     Description: The project name under which resources are created
+  oai:
+    Type: String
+    Description: The origin access identity ID associated with the environment
 
 Resources:
   Table:
@@ -66,13 +69,7 @@ Resources:
                   - !GetAtt Bucket.Arn
                   - "/*"
             Principal:
-              CanonicalUser: !GetAtt OriginAccessIdentity.S3CanonicalUserId
-
-  OriginAccessIdentity:
-    Type: AWS::CloudFront::CloudFrontOriginAccessIdentity
-    Properties:
-      CloudFrontOriginAccessIdentityConfig:
-        Comment: !Sub ${project}-cdn-${env}
+              AWS: !Sub "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity ${oai}"
 
 Outputs:
   Table:


configuration/exodus-storage.yaml
```diff
@@ -66,13 +69,7 @@ Resources:
...
                   - "/*"
             Principal:
-              CanonicalUser: !GetAtt OriginAccessIdentity.S3CanonicalUserId
-
-  OriginAccessIdentity:
-    Type: AWS::CloudFront::CloudFrontOriginAccessIdentity
-    Properties:
-      CloudFrontOriginAccessIdentityConfig:
-        Comment: !Sub ${project}-cdn-${env}
+              AWS: !Sub "arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity ${oai}"
```
rohanpm, line 72: I think this should probably be the full ARN, same as we did when parametrizing the roles.

Not sure where the "CloudFront Origin Access Identity" part of this comes from, but it seems human-oriented so probably is not wise to assume it'll always be exactly like that.  Also, if oai is not the full ARN, there is no reasonable way to figure out how to provide it into this template without reading the code and understanding "ah, I see, the string I provide is pasted onto the end of 'arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity ' so I should only provide what comes after that".

ARN is the natural language for identifying resources in AWS, it seems like that should be the default choice.

nathanegillett: So, `arn:aws:iam::cloudfront:user/CloudFront Origin Access Identity XXXXXXXXXX` is the full ARN but it's neither shown when creating the identity nor on [the identity console page](https://console.aws.amazon.com/cloudfront/home?region=us-east-1#oai:). It's **only** referenced in [this one documentation page](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html).

Also, the origin access identity attribute in the DistributionConfig (exodus-lambda-deploy.yaml) requires a different string, `origin-access-identity/cloudfront/XXXXXXXXXX` -- no alternatives.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-s3originconfig.html#cfn-cloudfront-distribution-s3originconfig-originaccessidentity

rohanpm: I see, thanks for explaining.

PR #87, opened 2020-06-29
remove OriginAccessIdentity in output

dichen16:


dichen16:
I found a test data needed to update,
```
https://cdn.redhat.com/content/dist/rhel/atomic/7/7Server/x86_64/ostree/repo/refs/heads/rhel-atomic-host/7/x86_64/standard: 
100%|█████████████████████████| 65.0/65.0 [00:00<00:00, 41.9kiB/s]

/content/dist/rhel/atomic/7/7Server/x86_64/ostree/repo/refs/heads/rhel-atomic-host/7/x86_64/standard verify checksum failed, 
cdn_data_checksum is f0299a2f60df6d60fbad8de6dfef4946c9302aae844fbc17051e9131a60acb14, 
but test_data_checksum is a7bd277f10209d6ddc6a6c086014c69f5a52d0fa6ab57cc817651de2b3f83e29
```
update here

support/reftest/data.yml
```diff
@@ -41,5 +41,5 @@ test_data:
   content-type: text/plain
 # */ostree/repo/refs/heads/*/*
 - path: /content/dist/rhel/atomic/7/7Server/x86_64/ostree/repo/refs/heads/rhel-atomic-host/7/x86_64/standard
-  sha256: a7bd277f10209d6ddc6a6c086014c69f5a52d0fa6ab57cc817651de2b3f83e29
+  sha256: f0299a2f60df6d60fbad8de6dfef4946c9302aae844fbc17051e9131a60acb14
```
rohanpm, line 44: Two points:

1. Can you please not update testdata like this within a PR which is completely unrelated to that.  Unrelated changes shouldn't be mixed together in a PR just because you happened to work on those changes around the same time.  Each PR should be focused on a particular topic.

2. If this happened once already it's probably going to happen again relatively frequently, so rather than just updating it you might want to look into if there's a better alternative.

dichen16: @rohanpm you are right, I will create separate PR.

PR #89, opened 2020-06-30
docs: add details on publishing tools

rohanpm:
We have a design for publishing tools now. Let's add them into the
architecture page so it's easier to see how everything fits
together.

exodus-gw, exodus-rsync repos don't exist yet; they should be
linked here once created.

It has been assumed that S3 uploads will be handled by exodus-gw,
but in truth this is still an open question; the diagram shows
the intended design and there is a chance that the plan will
change.


rohanpm:
https://imgur.com/HOFZNTh shows the rendered graph with this change.

PR #91, opened 2020-06-30
[RHELDST-1845] Skip the testdata checksum-verify of unstable content

dichen16:
1. reftest add unstable testdata check
2. the unstable content markded in data.yml

### DIFF
--- a/support/reftest/data.yml
+++ b/support/reftest/data.yml
@@ -39,7 +39,6 @@ test_data:
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/PULP_MANIFEST
   sha256: 2fb7a09fd67432ad7b7c95d24fe332827e24920347319c24bfef236cb0b72f19
   content-type: text/plain
-# */ostree/repo/refs/heads/*/*
+# */ostree/repo/refs/heads/*/* unstable content skip checksum-verify
 - path: /content/dist/rhel/atomic/7/7Server/x86_64/ostree/repo/refs/heads/rhel-atomic-host/7/x86_64/standard
-  sha256: a7bd277f10209d6ddc6a6c086014c69f5a52d0fa6ab57cc817651de2b3f83e29
   content-type: text/plain
--- a/support/reftest/reftest
+++ b/support/reftest/reftest
@@ -196,8 +196,8 @@ def prepare(db, s3, config, opt):
             url, opt.key, opt.cert, opt.cacert
         )
 
-        # checksum verify
-        if cdn_data_checksum != item["sha256"]:
+        # For unstable content which did not provide sha256, it will skip the checksum verify
+        if item.get("sha256") and cdn_data_checksum != item["sha256"]:
             print(
                 "{} verify checksum failed, cdn_data_checksum is {}, but test_data_checksum is {}".format(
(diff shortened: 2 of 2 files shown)


support/reftest/reftest
```diff
@@ -196,8 +196,13 @@ def prepare(db, s3, config, opt):
...
 
-        # checksum verify
-        if cdn_data_checksum != item["sha256"]:
+        # unstable content check
+        is_unstable = False
+        if not item["sha256"]:
+            is_unstable = True
+
+        # For unstable content, it will skip the checksum verify
+        if not is_unstable and cdn_data_checksum != item["sha256"]:
```
nathanegillett, line 205: `if not is_unstable` means that there **is** an item["sha256"]
Was this meant to be `if is_unstable`?
If not, that case must be handled or upload_from_localfile will fail.

dichen16: @nathanegillett  When `is_unstable == True`, the `not is_unstable` will be False, it won't go to the if block inside.
For unstable content don't need a checksum-verify.

nathanegillett: Ah, I see now that item["sha256"] below was changed to cdn_data_checksum. Nevermind.

On another note, this seems a little overcomplicated. You just want to ensure there is an item["sha256"] value before checking cdn_data_checksum, right? Why not this;
```suggestion
        # For unstable content, it will skip the checksum verify
        if item["sha256"] and cdn_data_checksum != item["sha256"]:
```
`is_unstable` isn't used anywhere else so it doesn't really seem necessary.

dichen16: @nathanegillett done!

support/reftest/reftest
```diff
@@ -196,8 +196,8 @@ def prepare(db, s3, config, opt):
             url, opt.key, opt.cert, opt.cacert
         )
 
-        # checksum verify
-        if cdn_data_checksum != item["sha256"]:
+        # For unstable content which did not provide sha256, it will skip the checksum verify
+        if item["sha256"] and cdn_data_checksum != item["sha256"]:
```
rohanpm, line 200: This is OK, but I think it's a little more friendly to make sha256 optional rather than forcing a blank entry in the file.  That could be done by using item.get('sha256') here.

support/reftest/data.yml
```diff
@@ -39,7 +39,7 @@ test_data:
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/PULP_MANIFEST
   sha256: 2fb7a09fd67432ad7b7c95d24fe332827e24920347319c24bfef236cb0b72f19
   content-type: text/plain
-# */ostree/repo/refs/heads/*/*
+# */ostree/repo/refs/heads/*/* unstable content skip checksum-verify
 - path: /content/dist/rhel/atomic/7/7Server/x86_64/ostree/repo/refs/heads/rhel-atomic-host/7/x86_64/standard
-  sha256: a7bd277f10209d6ddc6a6c086014c69f5a52d0fa6ab57cc817651de2b3f83e29
+  sha256:
```
rohanpm, line 44: Ideally you would also remove the sha256 from this line now.

The code makes it possible to omit sha256 for items now, but if the test data doesn't define at least one item without sha256, then there's no protection against this breaking.

PR #97, opened 2020-07-10
Remove versioning from CDN S3 buckets

### LINKED ISSUE #95: Disable versioning for CDN s3 buckets
#### Summary

Versioning for the CDN s3 bucket, configured in exodus-storage.yaml, is unnecessarily enabled.

#### Issue Type
- Bug Report

#### Steps to reproduce

n/a

#### Actual results

Versioning is enabled for CDN bucket

#### Expected results

Versioning is disabled for CDN bucket

#### OS / Environment

n/a

nathanegillett:
According to the current design, multiple object of the same key will
never be uploaded to the CDN S3 bucket, as the key is the sha256
checksum of the file. While probably harmless, leaving versioning
enabled can potentially hide duplicate uploads and drive up usage costs.

Closes #95

### DIFF
--- a/configuration/exodus-storage.yaml
+++ b/configuration/exodus-storage.yaml
@@ -50,8 +50,6 @@ Resources:
         BlockPublicPolicy: true
         IgnorePublicAcls: true
         RestrictPublicBuckets: true
-      VersioningConfiguration:
-        Status: Enabled
 
   BucketPolicy:
     Type: AWS::S3::BucketPolicy


configuration/exodus-storage.yaml
```diff
@@ -50,8 +50,6 @@ Resources:
         BlockPublicPolicy: true
         IgnorePublicAcls: true
         RestrictPublicBuckets: true
-      VersioningConfiguration:
```
rohanpm, line 53: In the case where the buckets already exist, does this actually disable versioning there or does it just leave the current setting alone?  Removing something from a config management tool isn't always enough to reverse the effect of previous runs.

nathanegillett: Right, buckets created prior to this change won't have versioning disabled by the template and would need to do that manually.

lioramilbaum: It worth checking if the s3_bucket Ansible module support it. If it does than we can think of way to execute it on existing bucket. WDYT?

nathanegillett: I don't think this something we need to worry about, as it won't be an issue going forward. I've already disabled versioning manually on exodus-cdn-dev/stage/prod buckets and developers can do what they like with their preexisting buckets.

nathanegillett:
@rohanpm's review is required for merge

PR #99, opened 2020-07-20
Update deployment documentation references to oai parameter

nathanegillett:


docs/deploy.rst
```diff
@@ -156,6 +157,8 @@ deployment, exodus-lambda-deploy.yaml, can be run by the `AWS CLI`_ command
 
 .. _AWS CLI: https://aws.amazon.com/cli/
 
+.. _AWS CoudFront: https://aws.amazon.com/cloudfront/
```
rohanpm, line 160:
```suggestion
.. _AWS CloudFront: https://aws.amazon.com/cloudfront/
```

docs/deploy.rst
```diff
@@ -22,6 +22,7 @@ pairs, exodus-storage.yaml, run by the `AWS CLI`_ command
         Default: exodus
     - oai:
         The origin access identity ID associated with the environment.
+        `AWS CloudFront`_ origin access identities are created separately.
```
rohanpm, line 25: It's not clear to me what this is trying to communicate to the reader.  Can it be spelled out in a little more detail?  Is it trying to say that, in order to find the value for this parameter, you should look at a cloudfront distribution and extract it from there?  Or trying to say that you should not share an oai between here and cloudfront?

nathanegillett: I guess this sentence isn't necessary as the parameter's existence implies that the origin access identity isn't created here, which is what I intended to convey.

nathanegillett: I'll rewrite this to be more helpful.

docs/deploy.rst
```diff
@@ -55,8 +55,7 @@ exodus-pipeline.yaml, run by the `AWS CLI`_ command
         Available values: dev, stage, prod
         Default: dev
     - oai:
-        The origin access identity ID associated with the environment,
-        created alongside the environment's bucket-table pair.
+        An `AWS CloudFront`_ origin access identity.
```
lioramilbaum, line 58: This value here should be the same one used for the storage step. Maybe adding another section for oai creation and than populating the value when running the other two steps. WDYT?

nathanegillett: Sorry, I don't think I follow. Are you referring to the sentence being repeated throughout the file?

lioramilbaum: The documentation is about to change, after the OAI resource creation is incorporated into the playbook. Having said that, my comment is not that important.

nathanegillett:
@rohanpm, this needs your approval

PR #101, opened 2020-07-22
Move exodus Content-Type handling from S3 to DynamoDB

lioramilbaum:


.gitignore
```diff
@@ -106,3 +106,6 @@ venv.bak/
 # deployment package
 package.zip
 package/
+
+# vscode
+.vscode
```
nathanegillett, line 111: I think it's preferred to list IDE-specific ignores to your global git ignore instead.

exodus_lambda/functions/origin_request.py
```diff
@@ -101,6 +100,10 @@ def handler(self, event, context):
                 request["uri"] = (
                     "/" + query_result["Items"][0]["object_key"]["S"]
                 )
+                if query_result["Items"][0]["content_type"]["S"]:
+                    request["uri"] += "&ResponseContentType={}".format(
```
nathanegillett, line 104: Content type should be separate from uri, i.e., `request["content_type"] = query_result["Items"][0]["content_type"]["S"]`.

.gitignore
rohanpm: I also use vscode but typically put the ignores into local ``.git/info/exclude``, I don't mind having this kind of thing added to the repo though.

docs/arch.gv
```diff
@@ -114,17 +119,14 @@ digraph {
                     <tr>
                         <td>aec070645fe5...</td>
                         <td><i>[blob some repomd.xml]</i></td>
-                        <td>{ContentType: application/xml}</td>
```
rohanpm, line 117: The table will be messed up now because these rows now only have 2 cells, while if you look at earlier rows in the same table, they have 3 cells, plus the header colspan is 3.

The other rows had ``-`` to denote "empty" for metadata, but now that there is always no metadata here, could we just delete the "metadata" column from all rows?

exodus_lambda/functions/origin_request.py
rohanpm: The comment about request["content_type"] doesn't seem right to me.

The goal here is to rewrite the request so that, if the DB says the response should be content-type X, then we arrange for S3 to respond with content-type X.  It can be achieved by including ResponseContentType in the query string before it goes to S3, documented at https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax (note: a little unclear from the docs if it's "response-content-type" or "ResponseContentType", guess it's the latter)

Setting request["content_type"] seems like it'd do something else... like overriding the content type of the incoming request from the user?  Or maybe it's invalid, I couldn't see any reference to content_type outside of headers at https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html#lambda-event-structure-request .

I think the overall approach here is right but the implementation seems slightly off:

- Adding & is only valid if there's already a non-empty query string, right?  Otherwise it should be ? to start the query string?
- the value put into the query string should be url-encoded.

Ideally we're using some library function to build the query string rather than doing it ad-hoc.

tests/functions/test_origin_request.py
```diff
@@ -81,7 +94,7 @@ def test_origin_request(
     assert "Item found for '%s'" % real_uri in caplog.text
 
     assert request == {
-        "uri": "/e4a3f2sum",
+        "uri": "/e4a3f2sum&ResponseContentType={}".format(content_type),
```
rohanpm, line 97: I think it's showing the wrong behavior:

- should be ? instead of &, right?
- the content types should be url-encoded

docs/conf.py
```diff
@@ -92,12 +92,12 @@
     "github_repo": "exodus-lambda",
     "github_button": False,
     "github_banner": True,
-    "description": "AWS Lambda functions for Red Hat's Content Delivery Network",
+    "description": "AWS Lambda functions for Red Hat's Content Delivery Network",  # noqa
```
rohanpm, line 95: Could this please be avoided?  I think this line is actually fine, so rather than adding noqa comments I'd rather fix the settings of whichever linter prompted you to add this.

I see in pyproject.toml right now we have a line length of 79 defined there, but I think it was a mistake, we're usually having a limit of 100.  Would you mind doing one of either:

- find/update/add the appropriate config file in this repo so your linter accepts line limit of 100, or
- just remove these noqa from the commit

.gitignore
lioramilbaum: I have moved it to the global git ignore.

exodus_lambda/functions/origin_request.py
rohanpm: About the URL encoding, what I meant is something like this:

```
                    query = urllib.parse.urlencode({'ResponseContentType': content_type})
                    request["uri"] += "?" + query
```

If content_type contains any special characters, they'll be encoded so that the other end can receive them properly, e.g. text/plain becomes text%2Fplain.

When I made the suggestion originally, I thought that it might be important to encode the `/` character.  But some of the APIs in urllib call it "safe", so now I'm not so sure.  That means I'm not 100% sure that any content types actually used will contain any characters which have to be encoded, but it's probably best to be on the safe side and encode anyway.

exodus_lambda/functions/origin_request.py
```diff
@@ -101,6 +101,14 @@ def handler(self, event, context):
                 request["uri"] = (
                     "/" + query_result["Items"][0]["object_key"]["S"]
                 )
+                if query_result["Items"][0]["content_type"]["S"]:
+                    content_type = query_result["Items"][0]["content_type"][
```
rohanpm, line 105: Guess this could be slightly cleaner if content_type is before the "if", that way you don't have to write the entire query_result[...] expression twice.

support/reftest/reftest
```diff
@@ -1,22 +1,22 @@
...
 Prepare reference test data for Exodus-lambda integration test.
 It requires a "data.yml" file specifying the reference test data.
 
 """
 
-import os
-import sys
-import yaml
-import boto3
+import argparse
```
dichen16, line 9: @lioramilbaum  small suggestion: could you pls change the import layout?
```
import sys
import yaml
import boto3
import hashlib
import tempfile
import argparse
import requests

from tqdm import tqdm
from datetime import datetime
```
it will look better.

lioramilbaum: isort is producing the current import layout, according to the best practices. Do we want to override those?

dichen16: @lioramilbaum I just checked the tool `isort`, it's pretty useful :+1:

exodus_lambda/functions/origin_request.py
nathanegillett: Ah, I clearly didn't know what's going on. Thanks for the doc link, @rohanpm.

lioramilbaum: The AWS documentation is confusing.  Let me double check.

rohanpm: That is pretty strange, but some examples show there isn't a literal `+`, e.g. there's an example like `GET /Junk3.txt?response-cache-control=No-cache`.

I have a feeling the `+` is meant to communicate something else, maybe the fact that `Key` is always at least one character.

The example now makes me question whether it should be ResponseContentType or response-content-type though (maybe both work?)

lioramilbaum: @rohanpm Is correct. The `+` is not required. Double checked that. In regards to the ResponseContentType vs. response-content-type I don't know the answer yet. Let's defer that to a later stage.

PR #105, opened 2020-08-03
Move exodus Content-Type handling from S3 to DynamoDB -  integration tests

lioramilbaum:
https://projects.engineering.redhat.com/browse/RHELDST-1324
https://projects.engineering.redhat.com/browse/RHELDST-2372


exodus_lambda/functions/origin_request.py
```diff
@@ -65,7 +64,9 @@ def handler(self, event, context):
         # pylint: disable=unused-argument
 
         request = event["Records"][0]["cf"]["request"]
+        self.logger.info("request is '%s'", request)
         uri = self.resolve_aliases(request["uri"])
+        self.logger.info("uri is '%s'", uri)
```
nathanegillett, line 69: Could you clean up the debugging messages, please?

A log message with this info might be helpful to keep around. So, if you want, you could convert them into a more user-friendly message.

tests/conftest.py
```diff
@@ -1,4 +1,7 @@
+#!/usr/bin/env python3
```
rohanpm, line 1: This is odd, why was this added?  The file isn't designed to be run.

exodus_lambda/functions/origin_request.py
```diff
@@ -103,11 +104,11 @@ def handler(self, event, context):
                 )
                 content_type = query_result["Items"][0]["content_type"]["S"]
                 if content_type:
-                    query = urllib.parse.urlencode(
-                        {"ResponseContentType": content_type}
-                    )
-                    request["uri"] += "?" + query
+                    request[
+                        "querystring"
+                    ] = "response-content-type={0}".format(content_type)
```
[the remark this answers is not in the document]
lioramilbaum: The encoding breaks the functionality. The tests are failing as soon as it is restored.

exodus_lambda/functions/origin_request.py
```diff
@@ -67,6 +66,13 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         uri = self.resolve_aliases(request["uri"])
 
+        self.logger.info(
+            "The request value for request_origin begining is '%s'", request
+        )
+        self.logger.info(
+            "The uri value for request_origin begining is '%s'", uri
+        )
```
nathanegillett, line 74: So, I think you meant `origin_request` instead of `request_origin`and it would be helpful to format the request dictionary a little to make it more readable; `json.dumps(request, indent=4, sort_keys=True)` should be good.

Similar changes would be welcome for the other log messages.
I don't think this is anything to block the PR over, though. We can follow up with another PR for logging improvements.

exodus_lambda/functions/origin_request.py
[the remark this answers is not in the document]
nathanegillett: If we're worried about slashes in the querystring...
> urllib.parse.quote(content_type, safe="")
'text%2Fplain'

But if integration tests pass without URL encoding, slashes _must_ be safe. I feel that AWS should accept encoded as well, though, so failures due to this don't make sense to me.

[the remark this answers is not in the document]
lioramilbaum: Are you sure encryption is required? I have reviewed the documentation again. Didn't find any reference to encryption for this one.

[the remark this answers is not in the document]
rohanpm: I'm not sure it's required.  I realized later that it may or may not be required because cloudfront might be decoding the query before it puts it into 'querystring', or it might not... and the docs are completely silent on the point.

It doesn't make sense to me that encoding wouldn't be needed in general though, seeing as it comes as a single string.  If you imagine for instance that the client wanted to pass the equivalent of `{'snacks': 'peanuts&chocolate', 'meals': 'sandwich'}` in their request, logically it *must* be coming into the lambda with that & encoded to %26, i.e. `meals=sandwich&snacks=peanuts%26chocolate`, otherwise there would be no safe way to use the query string.  So I conclude that either values in querystring *are* meant to be encoded, or else the lambda event design is majorly flawed.

Anyway, due to lack of details in the docs, it seems we just need to accept whatever works.

exodus_lambda/functions/origin_request.py
```diff
@@ -67,6 +66,14 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         uri = self.resolve_aliases(request["uri"])
 
+        self.logger.info(
+            "The request value for origin_request begining is '%s'",
```
rohanpm, line 70:
```suggestion
            "The request value for origin_request beginning is '%s'",
```

exodus_lambda/functions/origin_response.py
```diff
@@ -14,6 +15,15 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         response = event["Records"][0]["cf"]["response"]
 
+        self.logger.info(
+            "The request value for origin_response begining is '%s'",
```
rohanpm, line 19:
```suggestion
            "The request value for origin_response beginning is '%s'",
```

exodus_lambda/functions/origin_response.py
```diff
@@ -14,6 +15,15 @@ def handler(self, event, context):
...
+            json.dumps(request, indent=4, sort_keys=True),
+        )
+        self.logger.info(
+            "The response value for origin_response begning is '%s'",
```
rohanpm, line 23:
```suggestion
            "The response value for origin_response beginning is '%s'",
```

exodus_lambda/functions/origin_request.py
```diff
@@ -67,6 +66,14 @@ def handler(self, event, context):
         request = event["Records"][0]["cf"]["request"]
         uri = self.resolve_aliases(request["uri"])
 
+        self.logger.info(
```
rohanpm, line 69: Anyone want to share some thoughts on whether INFO or DEBUG is appropriate here?

We should probably have some rough guidelines to tell us when a log message might occur too often or be too verbose for INFO level.  Maybe a log message which occurs once per request (like these) is fine for INFO, while a log message which occurs many times per request (e.g. in a loop) would be better as DEBUG.

lioramilbaum:
I am breaking this PR into smaller PRs. The purpose is to increase the lead time. Please review #106.

exodus_lambda/functions/origin_request.py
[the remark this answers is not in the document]
nathanegillett: Yeah, I think the reason encoding the `content_type` right now doesn't seem necessary is because only strings containing slashes are tested. Slashes may be safe, then, but we can still encode the `content_type` to protect against other special characters that may be introduced in the future.
```suggestion
                    ] = "response-content-type={0}".format(urllib.parse.quote(content_type))
```
This shouldn't change the current behavior, "text/plain" is still "text/plain".

exodus_lambda/functions/origin_request.py
nathanegillett: I like that guideline, myself. These do make enough sense to me as INFO messages.

nathanegillett:
So, @lioramilbaum, is this PR being closed in favor of #106 and perhaps another PR for logging?

lioramilbaum:
@nathanegillett Correct. I will close this PR as soon as I file the logging PR.

lioramilbaum:
Second PR #107

PR #106, opened 2020-08-05
Move exodus Content-Type handling from S3 to DynamoDB - integration tests

lioramilbaum:
The previous behavior of adding query string to uri was broken.
querystring should be used instead and use the 'response-content-type'
attriute.

https://projects.engineering.redhat.com/browse/RHELDST-1324
https://projects.engineering.redhat.com/browse/RHELDST-2372

I am breaking down PR #105 to smaller ones to increase the lead time.


exodus_lambda/functions/origin_request.py
```diff
@@ -103,10 +102,9 @@ def handler(self, event, context):
                 )
                 content_type = query_result["Items"][0]["content_type"]["S"]
                 if content_type:
-                    query = urllib.parse.urlencode(
-                        {"ResponseContentType": content_type}
-                    )
-                    request["uri"] += "?" + query
+                    request[
+                        "querystring"
+                    ] = "response-content-type={0}".format(content_type)
```
nathanegillett, line 107: As mentioned here: https://github.com/release-engineering/exodus-lambda/pull/105#discussion_r465682860
If we want to protect against special characters aside from "/";
```suggestion
                    request[
                        "querystring"
                    ] = "response-content-type=%s" % urllib.parse.quote(content_type)
```

But if slashes are the only special character ever used here, then I guess there's no reason for encoding.

At the very least, though, I'd suggest we use the older "%" means of interpolation, just to be consistent with what's already used throughout the repository.
```suggestion
                    request[
                        "querystring"
                    ] = "response-content-type=%s" % content_type
```

exodus_lambda/functions/origin_request.py
```diff
@@ -103,10 +102,9 @@ def handler(self, event, context):
                 )
                 content_type = query_result["Items"][0]["content_type"]["S"]
                 if content_type:
-                    query = urllib.parse.urlencode(
-                        {"ResponseContentType": content_type}
+                    request["querystring"] = (
+                        "response-content-type=%s" % content_type
```
rohanpm, line 106: Sorry to keep on the same issue here, but the lack of encoding here still doesn't sit right with me.

Per the earlier discussion, I concluded that encoding is needed in querystring *in general* if we're dealing with arbitrary strings.

We aren't dealing with arbitrary strings, so it comes down to whether or not any content type might contain a character which needs encoding.  I went searching for relevant standards and couldn't convince myself that it's safe not to encode.  I found https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

The standard says (roughly) that content types can contain any ASCII except one of:

```
()<>@,;:\"[]?.=
```

Notably: `%` is not included in the list of banned characters there, so a content type containing a literal `%` is apparently possible. If that happens then it doesn't seem likely to be interpreted by S3 correctly if it was not encoded.  I tried testing it but it's a little cumbersome.

Could you please have another try of making this encoded, so we are not laying a trap for our future selves?

You mentioned earlier that it causes tests to fail, if you still find this to be the case could you please share the code when it failed, and the output from the tests?

PR #109, opened 2020-08-11
[RHELDST-2414] integration tests should log response headers

dichen16:
In exodus-lambda tests/integration, when a test fails because some attribute of
a response doesn't match an expected value, the test output contains all headers
on the response.

### DIFF
--- a/tests/integration/test_exodus.py
+++ b/tests/integration/test_exodus.py
@@ -1,3 +1,4 @@
+import json
 import pytest
 import requests
 
@@ -9,6 +10,7 @@ def test_exodus_basic(cdn_test_url):
     )
 
     r = requests.get(url)
+    print(json.dumps(dict(r.headers), indent=2))
     assert r.status_code == 200
     assert "cache-control" not in r.headers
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_exodus.py
```diff
@@ -9,6 +9,7 @@ def test_exodus_basic(cdn_test_url):
     )
 
     r = requests.get(url)
+    print(r.headers)
```
lioramilbaum, line 12: How about formatting the json in an easier way to read it. 
https://www.journaldev.com/33302/python-pretty-print-json

dichen16: @lioramilbaum Yes! I will give it a try.

dichen16: @lioramilbaum done!

dichen16:
@rohanpm @nathanegillett @crungehottman @lioramilbaum 
The reason why I did the type casting `dict(r.headers)` is if it use r.headers directly will encounter the following problem.
```
TypeError: Object of type CaseInsensitiveDict is not JSON serializable
```
and the type of `r.headers` is `<class 'requests.structures.CaseInsensitiveDict'>`.
This issue can also explain it. https://github.com/psf/requests/issues/1380

dichen16:
@rohanpm updated the commit message.

PR #116, opened 2021-01-27
exodus-lambda requires cdn-definitions-private [RHELDST-4703]

crungehottman:
The configuration that was once available in the cdn-definitions
PyPI package has since moved to a private git repository:
"cdn-definitions-private".

In order for exodus-lambda to consume data from a private repo,
a "GITHUB_TOKEN" environment variable must be set. GITHUB_TOKEN
should be assigned a GitHub access token with the "repo" scope,
which enables clients (e.g., pip) access to private repos.

### DIFF
--- a/scripts/build-package
+++ b/scripts/build-package
@@ -3,6 +3,8 @@ set -e
 
 pip install --require-hashes -r requirements.txt --target ./package
 pip install --no-deps --target ./package .
+git clone git+https://${GITHUB_TOKEN}@github.com/release-engineering/cdn-definitions-private.git
+mv ./cdn-definitions-private/data.yaml ./package/cdn_definitions/data.yaml
 cp ./configuration/exodus-lambda-deploy.yaml ./package
 envsubst < ./configuration/lambda_config.json > ./package/lambda_config.json
 aws cloudformation package \
--- a/configuration/exodus-pipeline.yaml
+++ b/configuration/exodus-pipeline.yaml
@@ -221,8 +221,11 @@ Resources:
                 Region: !Ref region
                 Configuration:
                   EnvironmentVariables:
-                    !Sub '[{"name": "ENV_TYPE", "value": "${env}", "type": "PLAINTEXT"},
-                    {"name": "PROJECT", "value": "${project}", "type": "PLAINTEXT"}]'
+                    !Sub '[
+                      {"name": "ENV_TYPE", "value": "${env}", "type": "PLAINTEXT"},
+                      {"name": "PROJECT", "value": "${project}", "type": "PLAINTEXT"},
+                      {"name": "GITHUB_TOKEN", "value": "${githubToken}", "type": "PLAINTEXT"}
+                    ]'
                   ProjectName: !Sub ${project}-lambda-build
                 InputArtifacts:
                   - Name: SourceArtifact


tox.ini
```diff
@@ -25,13 +29,17 @@ commands=
     pip install --require-hashes -r requirements.txt --target ./package
     pip install --no-deps --target ./package .
 
+    # Install packages from private repositories
+    pip install --target ./package git+https://{env:GITHUB_TOKEN:}@github.com/release-engineering/cdn-definitions-private.git#egg=cdn_definitions_private
```
rohanpm, line 33: Can we please add --no-deps here as well?  The goal is to ensure there's no untrusted installs of indirect dependencies (e.g. if cdn-definitions-private declared a dependency on something which didn't appear in requirements.txt of this repo).

rohanpm (commented):
Looks OK, no major concerns here.

nathanegillett (dismissed):

scripts/build-package
```diff
@@ -2,6 +2,8 @@
 set -e
 
 pip install --require-hashes -r requirements.txt --target ./package
+rm -rf ./package/cdn_definitions*
```
rohanpm, line 5: The implication here seems to be that if you want to use the private data, you're supposed to uninstall cdn-definitions.  So you can have cdn-definitions or cdn-definitions-private installed, but not both.  Is that right?

If so - I don't think it should be that way.  I think it should be more like: if you're going to use any cdn-definitions dataset, then you install cdn-definitions and that gives you reference data, plus functionality common to any dataset, such as the expected schema for the dataset and maybe a handful of utility functions.  Then you can install cdn-definitions-private to *supplement* that with the desired private dataset - not *replace* it.  If you have to uninstall cdn-definitions to install cdn-definitions-private then that implies you lose the common functionality (such as the schema) or that cdn-definitions-private has to duplicate it.

Should we maybe work on a brief design doc explaining a bit more the scope of each project?  Seems like it might not be entirely clear.

rohanpm (changes requested):

scripts/build-package
```diff
@@ -3,6 +3,10 @@ set -e
 
 pip install --require-hashes -r requirements.txt --target ./package
 pip install --no-deps --target ./package .
+git clone git+https://${GITHUB_TOKEN}@github.com/release-engineering/cdn-definitions-private.git
+./cdn-definitions-private/scripts/make-json
```
crungehottman, line 7: depends on https://gitlab.cee.redhat.com/cdn/cdn-definitions-private/-/merge_requests/6

nathanegillett (dismissed):

scripts/build-package
rohanpm: I don't think we should do this:

The approach being proposed here, of having cdn-definitions-private only keep a YAML file and then running a script to produce a JSON file, doesn't make sense to me.  I've written a bigger comment at https://github.com/release-engineering/cdn-definitions/pull/19/files#r568302386 about YAML vs JSON.

I think it would make sense to focus on completion of the cdn-definitions change first and then come back to this once that's been released.

rohanpm (changes requested):

nathanegillett (approved):

nathanegillett (approved):

scripts/build-package
```diff
@@ -3,6 +3,8 @@ set -e
 
 pip install --require-hashes -r requirements.txt --target ./package
 pip install --no-deps --target ./package .
+git clone git+https://${GITHUB_TOKEN}@github.com/release-engineering/cdn-definitions-private.git
```
lipoja, line 6: Since I do not have the overall knowledge of this I am going to ask this question:
Will be the directory with cloned repo deleted before next run of this script? Right now when you run this multiple times it will not fetch latest changes, it just says directory already exists and is not empty.

lipoja: It looks like it might be run on AWS pipeline, and I guess it starts always with clear data storage? So you will not encounter that issue mentioned above. Am I right?

crungehottman: > So you will not encounter that issue mentioned above. Am I right?

Yes, you are correct. This script is intended to be run by AWS CodeBuild. CodeBuild spins up a fresh build environment for each build.

rohanpm (approved):

PR #118, opened 2021-03-04
The bump to cdn-definitions-2.0.0 requires additional dependencies

crungehottman:
The CI in https://github.com/release-engineering/exodus-lambda/pull/117 will continue to fail without this commit.


nathanegillett (approved):

nathanegillett:
Q: Could these requirement changes not be made in the same commit as the cdn-definitions bump?

rohanpm:
> Q: Could these requirement changes not be made in the same commit as the cdn-definitions bump?

Dependabot in this repo is unable to handle that because we haven't added separate ".in", ".txt" files, e.g. (requirements.in, requirements.txt).  The dependabot behavior in this mode is that it'll only update direct dependencies one by one without considering indirect dependencies, and that's incompatible with hash-checking mode, hence breaking CI.

If we update this project to use .in files like others, then dependabot will use pip-compile and it should all work.

rohanpm (approved):

PR #148, opened 2021-03-16
Add missing test-requirement causing --require-hashes error

nathanegillett:
The typing-extensions package, a dependency of importlib-metadata which
is a dependency of pytest, doesn't have it's version pinned in
test-requirements.txt.


nathanegillett:
I created this while referencing the wrong CodeBuild.

crungehottman (dismissed):

nathanegillett:
Updated the commit message, as this error seems to be present at all levels.

crungehottman (approved):

PR #160, opened 2021-04-01
Update testing and dependency tools

nathanegillett:
This commit includes the addition of isort and mypy checks, transition
to pip-compile requirements listing, and shifts to Python3.7 for
testing, as exodus-lambda builds and runs using that version.

The isort and black tools were run against the repository and some test
changes were necessary upon switching to pip-compile dependencies.


nathanegillett:
This brings exodus-lambda inline with exodus-gw for testing/workflow stuff.

crungehottman (approved):
:clap: :clap: :clap:

PR #167, opened 2021-04-07
Move CI to Github actions from Travis.ci [RHELDST-5504]

nathanegillett:
Additionally introduces linting action.


nathanegillett:
Unsure why Travis CI check is still present or why coveralls is having issues...

nathanegillett:
Moving back to the coveralls Python library instead of a coveralls Github action has fixed the issue in which coverage was incorrectly reported.

Apologies for the push spam while I worked that out.

PR #284, opened 2021-10-25
Consume cdn-definitions from table [RHELDST-7898]

crungehottman:
The exodus-lambda functions previously consumed the cdn definitions
from the the exodus-lambda package. This workflow required the lambdas
to be rebuilt and redeployed after each update to the cdn-definitions
data set.

Release engineers should be able to update the cdn-definitions data
set without rebuilding and redeploying the lambdas. The definitions
are now stored in a $project-config-$env DynamoDB table, which is
decoupled from the exodus-lambda package.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -1,27 +1,53 @@
 import json
+import time
 import urllib
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 import boto3
-from cdn_definitions import load_data
+import cachetools
 
 from .base import LambdaBase
 
 
 class OriginRequest(LambdaBase):
-    def __init__(
-        self, conf_file="lambda_config.json", definitions_source=None
-    ):
+    def __init__(self, conf_file="lambda_config.json"):
         super().__init__("origin-request", conf_file)
         self._db_client = None
-        self._definitions_source = definitions_source
-        self._definitions = None
+        self._cache = cachetools.TTLCache(
+            maxsize=1,
+            ttl=timedelta(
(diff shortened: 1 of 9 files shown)


nathanegillett (dismissed):

exodus_lambda/functions/origin_request.py
```diff
@@ -1,27 +1,52 @@
...
+                                timespec="milliseconds"
+                            )
+                        )
+                    },
+                },
+            )
+            if query_result["Items"]:
+                item = query_result["Items"][0]
+                self._cache["exodus-config"] = json.loads(item["config"]["S"])
+        return self._cache.get("exodus-config")
```
rohanpm, line 49: Minor: it would generally be a bad practice to write code like this when using a cache from cachetools:

```
cache["some-key"] = ...
return cache.get("some-key")
```

Problem: you don't know if the cache has become invalidated again between line 1 and 2.  The cache invalidates itself without warning.  So, for example, for a time-based cache it can expire between 1 and 2 due to the passage of time; for a cache with a maxsize it can expire between 1 and 2 because another thread ran meanwhile and inserted something else into the cache bumping it past the maxsize and causing evictions.  Just because you put something in the cache at one line of code doesn't guarantee that it's still present when the next line of code executes.

The safe way to write the code would be more like:

```
out = ...
cache["some-key"] = out
return out
```

So you write to the cache, but then you don't assume that the value you just wrote can be pulled back out.

I'm calling it minor since the specific choice of cache, and absence of threads in this code should make it very unlikely for any genuine issue here.

rohanpm (approved):

nathanegillett (approved):

rohanpm (approved):

PR #303, opened 2021-12-01
exodus-lambda crashed when missing content-type in DynamoDB [RHELDST-…

dichen16:
…8905]

When a client requests an item with no content_type set, the response
should include a Content-Type of "application/octet-stream".

Add a new test case cover the empty content_type scenario.

### DIFF
--- a/support/reftest/reftest
+++ b/support/reftest/reftest
@@ -38,15 +38,20 @@ class DBHandler:
     ):
         if not from_date:
             from_date = self.default_from_date
-        return self.dynamodb.put_item(
+
+        item = {
+            "web_uri": {"S": web_uri},
+            "from_date": {"S": from_date},
+            "object_key": {"S": object_key},
+            "metadata": {"M": metadata},
+        }
+
+        if content_type:
+            item["content_type"] = {"S": content_type}
+
+        self.dynamodb.put_item(
             TableName=self.table,
-            Item={
-                "web_uri": {"S": web_uri},
-                "from_date": {"S": from_date},
-                "object_key": {"S": object_key},
-                "content_type": {"S": content_type},
-                "metadata": {"M": metadata},
-            },
+            Item=item,
         )
 
 
@@ -213,7 +218,7 @@ def prepare(db_client, s3_client, config, opt):
             item["path"],
             cdn_data_checksum,
             from_date=opt.release_date,
-            content_type=item["content-type"],
+            content_type=item.get("content-type", None),
         )
 
         # delete the NamedTemporaryFile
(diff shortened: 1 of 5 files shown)


support/reftest/reftest
```diff
@@ -38,7 +38,9 @@ class DBHandler:
     ):
         if not from_date:
             from_date = self.default_from_date
-        return self.dynamodb.put_item(
+        if not content_type:
+            content_type = ""
```
rohanpm, line 42: If the intent here was to cover what happens when there's no content_type, I don't think this quite achieves that, does it?  This will store content_type of `""` (empty string) on the created items.

Most items in the DB right now have *no* content_type at all, which is slightly different.  Should you do the same here?

tests/functions/test_origin_request.py
```diff
@@ -74,6 +79,7 @@
         "no alias keywords",
         "releasever alias",
         "layered rhui, releasever alias",
+        "no content_type"
```
rohanpm, line 82: Similarly as the comment against reftest, this does not cover the realistic case properly.  Realistically, items can have no content_type at all which is different from having a content_type of an empty string.  In fact the old code would already not crash if an item had content_type of `""`, so adding a test for that value is not going to prove that the crash was fixed.

I think you ought to edit the test (or add a new one) which truly tests the behavior on *no* content_type, rather than a content_type of `""`.

rohanpm (changes requested):

support/reftest/reftest
```diff
@@ -38,15 +38,24 @@ class DBHandler:
     ):
         if not from_date:
             from_date = self.default_from_date
-        return self.dynamodb.put_item(
-            TableName=self.table,
-            Item={
+        if not content_type:
+            item = {
```
rohanpm, line 42: Seems like it'd be better to not duplicate the setting of most of the item fields, since that invites bugs where someone updates one path and forgets the other.

Would it make more sense to write it like this:

```python
    item = {
       "web_uri": # ...,
       # ...rest of fields except content_type
    }
    if content_type:
        item["content_type"] = {"S": content_type}
```

So that there are not two places where web_uri, object_key etc are being set.

rohanpm (approved):

tests/functions/test_origin_request.py
```diff
@@ -221,3 +221,60 @@ def test_origin_request_definitions_cache(
...
+        ),
+    ],
+    ids=[
+        "no content_type",
+    ],
+)
+@mock.patch("boto3.client")
+@mock.patch("exodus_lambda.functions.origin_request.cachetools")
+@mock.patch("exodus_lambda.functions.origin_request.datetime")
+def test_origin_request_no_content_type(
```
nathanegillett, line 241: Minor:
Setup for this test seems odd to me. It's using parametrize but only has one case and in that case both parameters are identical. Maybe it's meant to be expanded later? Though, we probably should only build things around what's strictly necessary rather than future potential.

nathanegillett (approved):

crungehottman (approved):

PR #305, opened 2021-12-06
Generate listing responses [RHELDST-7949]

nathanegillett:
Using the cached definitions from the exodus-config DynamoDB table,
requests for content listings are now served directly from exodus-lambda
in the HTTP response body instead of querying for and downloading actual
file-objects.


nathanegillett:
> $ curl https://d3fn7bx681wuw4.cloudfront.net/content/dist/rhel/client/6/listing
> 6.1
> 6.10
> 6.2
> 6.3
> 6.4
> 6.5
> 6.6
> 6.7
> 6.8
> 6.9
> 6Client

(compare to http://cdn.stage.redhat.com/content/dist/rhel/client/6/listing)

exodus_lambda/functions/origin_request.py
```diff
@@ -100,11 +100,37 @@ def resolve_aliases(self, uri):
...
+            "statusDescription": "OK",
+        }
+
     def handler(self, event, context):
         # pylint: disable=unused-argument
 
         request = event["Records"][0]["cf"]["request"]
         uri = self.resolve_aliases(request["uri"])
+
+        if uri.endswith("/listing"):
```
rohanpm, line 131: I know we've had discussion about this before and I think we said that the lambda code should be the only way of handling a request for listing files.  However, now seeing it in the code it made me nervous as we are essentially reserving the filename "listing" here, to mean only this kind of file and making it impossible to use for anything else.

I wondered are we really sure that the 'listing' filename is unused for other purposes, even for "raw rsynced" directory trees like kickstarts, so I did some investigation.  Turns out the answer is: they are sometimes used for other things.

An example is that I found /content/dist/rhel/server/6/6.1/x86_64/kickstart/ResilientStorage/listing, which exists and has a list of RPMs rather than $releasever or $basearch values.

So, it looks like we can't unconditionally assume that all URLs ending with "/listing" need to be handled by this mechanism as we would then block the publish of files named "listing" for other purposes, including the existing content above.

Could we please change it so that, if lookup_listing didn't find an entry for that listing file, we proceed with the normal handling?

rohanpm (changes requested):

rohanpm (approved):

PR #306, opened 2021-12-13
exodus-cdn-playbooks should be able to configure log levels & more [R…

dichen16:
…HELDST-9223]

### DIFF
--- a/configuration/lambda_config.json
+++ b/configuration/lambda_config.json
@@ -11,9 +11,9 @@
       "us-east-1"
     ]
   },
-  "config_cache_ttl": 2,
+  "config_cache_ttl": $EXODUS_CONFIG_CACHE_TTL,
   "headers": {
-    "max_age": "600"
+    "max_age": "$EXODUS_HEADERS_MAX_AGE"
   },
   "logging": {
     "version": 1,
@@ -27,10 +27,10 @@
     },
     "loggers": {
       "origin-response": {
-        "level": "WARNING"
+        "level": "$ORIGIN_RESPONSE_LOGGER_LEVEL"
       },
       "origin-request": {
-        "level": "WARNING"
+        "level": "$ORIGIN_REQUEST_LOGGER_LEVEL"
       },
       "default": {
         "level": "WARNING"
--- a/exodus_lambda/functions/base.py
+++ b/exodus_lambda/functions/base.py
@@ -18,22 +18,8 @@ class LambdaBase(object):
             if isinstance(self._conf_file, dict):
                 self._conf = self._conf_file
             else:
-                for conf_file in [
-                    self._conf_file,
-                    os.path.join(
-                        os.path.dirname(
-                            os.path.dirname(os.path.dirname(__file__))
-                        ),
-                        "configuration",
-                        "lambda_config.json",
-                    ),
-                ]:
-                    if os.path.exists(conf_file):
-                        with open(
-                            conf_file, "r", encoding="UTF-8"
-                        ) as json_file:
-                            self._conf = json.load(json_file)
-                        break
+                with open(self._conf_file, "r", encoding="UTF-8") as json_file:
+                    self._conf = json.load(json_file)
         return self._conf
 
     @property
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -1,4 +1,5 @@
 import json
+import os
 import time
 import urllib
 from datetime import datetime, timedelta, timezone
@@ -8,9 +9,11 @@ import cachetools
 
 from .base import LambdaBase
 
+CONF_FILE = os.environ.get("EXODUS_LAMBDA_CONF_FILE") or "lambda_config.json"
+
 
 class OriginRequest(LambdaBase):
-    def __init__(self, conf_file="lambda_config.json"):
+    def __init__(self, conf_file=CONF_FILE):
         super().__init__("origin-request", conf_file)
         self._db_client = None
         self._cache = cachetools.TTLCache(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,7 @@
+import os
+import subprocess
+import tempfile
+
 import boto3
 import pytest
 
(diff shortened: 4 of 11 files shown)


configuration/lambda_config.json
```diff
@@ -27,10 +27,10 @@
     },
     "loggers": {
       "origin-response": {
-        "level": "WARNING"
+        "level": "$ORIGIN_RESPONSE_LOGGER_LEVEL"
```
rohanpm, line 30: You should have some default values for these vars.  If you don't, then this new version of the code can't be deployed by the current version of the playbooks & pipeline.

I'd suggest doing it like this in `scripts/build-package` :

```
export ORIGIN_RESPONSE_LOGGER_LEVEL=${ORIGIN_RESPONSE_LOGGER_LEVEL:-WARNING}
export ORIGIN_REQUEST_LOGGER_LEVEL=${ORIGIN_REQUEST_LOGGER_LEVEL:-WARNING}
```

That will set the env vars to WARNING if and only if they are not already set.  Must be done before the call to 'envsubst'.

tests/functions/test_origin_response.py
```diff
@@ -60,7 +60,7 @@ def test_origin_response_valid_headers(original_uri, want_digest):
             }
         ]
 
-    response = OriginResponse(conf_file=CONF_PATH).handler(event, context=None)
+    response = OriginResponse(conf_file=TEST_CONF).handler(event, context=None)
```
negillett, line 63: The CONF_PATH => TEST_CONF change isn't necessary but doesn't hurt anything either.

negillett (dismissed):

negillett:
This LGTM as is, provided the exodus-cdn-playbooks change with default values is submitted and deployed first.

tests/test_utils/utils.py
```diff
@@ -1,42 +1,55 @@
...
-        "format"
-    ] = "[%(levelname)s] - %(message)s\n"
-    conf["logging"]["loggers"]["origin-response"]["level"] = "DEBUG"
-    conf["logging"]["loggers"]["origin-request"]["level"] = "DEBUG"
-    conf["logging"]["loggers"]["default"]["level"] = "DEBUG"
-
-    conf["config_table"]["name"] = "test-config-table"
-    conf["config_table"]["cache_ttl"] = 2
 
+def generate_test_config(conf="configuration/lambda_config.json"):
```
negillett, line 7: This doesn't seem very useful now that we're setting up a fake config file in conftest. Shouldn't we just be using that for all tests?

The only difference seems to be the table's available_regions. We can just add the 5 region names to the conftest config.

exodus_lambda/functions/base.py
```diff
@@ -18,22 +18,12 @@ def conf(self):
...
-                    ),
-                ]:
-                    if os.path.exists(conf_file):
-                        with open(
-                            conf_file, "r", encoding="UTF-8"
-                        ) as json_file:
-                            self._conf = json.load(json_file)
-                        break
+                # to be polished: what if the self._conf_file not exist
+                if os.path.exists(self._conf_file):
```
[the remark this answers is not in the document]
negillett: Hmm. Maybe it does make sense to remove that default, since the config in that file isn't usable as-is.
In the case self._conf_file doesn't exist, maybe it's best to just exit/fail rather than assume any details about the project.

exodus_lambda/functions/origin_request.py
```diff
@@ -8,9 +9,11 @@
 
 from .base import LambdaBase
 
+CONF_FILE = os.environ.get("EXODUS_LAMBDA_CONF_FILE") or "lambda_config.json"
```
crungehottman, line 12:
```suggestion
CONF_FILE = os.environ.get("EXODUS_LAMBDA_CONF_FILE", "lambda_config.json")
```

rohanpm: In general I prefer the former style as Tom has written it now.

The reason is that it allows you to set an empty string for the env var in order to mean "use the default behavior" (while with your suggestion, setting an empty string would just make it crash).

This is useful because sometimes, when various systems are plugged together, you might find yourself in a situation where you can't possibly unset EXODUS_LAMBDA_CONF_FILE, but you *can* set it to an empty string.

For instance, imagine it's coming from an ansible variable filled in when you run a playbook, and there's a default value in inventory already.  If you run ansible at the command-line, you can then do `-e exodus_lambda_conf_file=""` to set a variable to an empty string in order to mean "just use the default".  AFAIK you can't *unset* a variable which has already been set though.

Therefore as a general rule I think it's a good idea to treat empty values the same as absent values when reading config, because sometimes you have a way to make a value empty, but you don't have a way to make it absent.

tests/conftest.py
```diff
@@ -41,3 +45,29 @@ def cdn_test_url(request):
     else:
         pytest.skip("Test skipped without --cdn-test-url or --lambda-stack")
     return url
+
+
+def mock_conf_file():
+    temp_file = tempfile.NamedTemporaryFile(delete=False)
```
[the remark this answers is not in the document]
rohanpm: Also please use a 'prefix' or 'suffix'.  It's a good practice so that if /tmp fills up with files, you can get some idea of which code created those files.

tests/conftest.py
```diff
@@ -41,3 +45,29 @@ def cdn_test_url(request):
...
+
+    os.environ["PROJECT"] = "test"
```
rohanpm, line 53: Hmm, I don't think you should dump these into the global os.environ here.  Note these stay in place for the whole test.

I think this generates confusion as the code under test is going to see env vars such as ORIGIN_RESPONSE_LOGGER_LEVEL which will never be there in production.  Someone might be wrongly led to believe that these env vars are present when the code runs, and even write code which depends on that with the tests wrongly passing.

subprocess.run takes an `env` argument which sets the environment for just that process, you should probably be using that so the env vars don't pollute all the tests.

EXODUS_LAMBDA_CONF_FILE is the only exception to this as that one really needs to be set for the code to work.

exodus_lambda/functions/base.py
```diff
@@ -18,22 +18,12 @@ def conf(self):
...
-                        "lambda_config.json",
-                    ),
-                ]:
-                    if os.path.exists(conf_file):
-                        with open(
-                            conf_file, "r", encoding="UTF-8"
-                        ) as json_file:
-                            self._conf = json.load(json_file)
-                        break
+                # to be polished: what if the self._conf_file not exist
```
rohanpm, line 21: How about just don't do the os.path.exists check?  If the config file doesn't exist, just let the open() call crash - what else can you do?  If you added custom error handling, you'd just be obscuring the error.

exodus_lambda/functions/base.py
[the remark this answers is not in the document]
rohanpm: IMO it makes sense to remove it but I thought we were going to rename the file as well to make it even clearer.

The reality is that the lambda_config.json checked into the repo is NOT a config file, it's a template used to generate a config file.  I think the JIRA mentioned renaming the file at one point to make it clear, but did the idea get lost?  It still makes sense to me and then it'd be absolutely obvious that trying to directly load the file as config is wrong.

rohanpm (changes requested):

tests/conftest.py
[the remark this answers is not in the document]
dichen16:
```
temp_file = tempfile.NamedTemporaryFile(prefix="lambda_unitests_", delete=False)
```
will create tempfile, like "/tmp/lambda_unitests_jypgdwwz", seems OK.

tests/test_utils/utils.py
dichen16: +1, yeah, keep the conf consistent.

dichen16:
I made the following changes.

1. clean the temporary file after test finished
2. remove the os.path.exists check
2. name the temporary file with a prefix indicating the usage
3. make subprocess.run take an env argument which sets the environment
4. make generate_test_config use the envsubst-ed config

CC @negillett @rohanpm @crungehottman

tests/conftest.py
```diff
@@ -41,3 +45,43 @@ def cdn_test_url(request):
...
+    )
+
+    return temp_file
+
+
+TEMP_FILE = mock_conf_file()
+
+
+def pytest_sessionfinish(session, exitstatus):
+    # remove temp conf after whole test run finished
```
rohanpm, line 85: Minor - instead of creating the file with delete=False and then having to both close and delete it here, how about removing the delete=False ?  Then you just need to close the file and it will be deleted.

tests/conftest.py
```diff
@@ -41,3 +45,43 @@ def cdn_test_url(request):
...
+        stderr=subprocess.STDOUT,
+        check=True,
+        shell=True,
+        env=test_env,
+    )
+
+    return temp_file
+
+
+TEMP_FILE = mock_conf_file()
```
rohanpm, line 81: Minor - TEMP_FILE is not a great name for this since there could easily be a need to have some other temp file later on as well.

rohanpm (approved):

dichen16:
[Update]
1. Polish `TEMP_FILE` to `MOCK_LAMBDA_CONF_FILE`
2. create tempfile with delete=True(default), the file is deleted as soon as it is closed. 
@rohanpm resolved two minor suggestion.

rohanpm (approved):

PR #307, opened 2021-12-14
Update Python 3.7 -> 3.9 for tests

negillett:
We expect exodus-lambda to be built and run with Python 3.9, so we
should be testing with the same version.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37, static, docs
+envlist = py39, static, docs
 
 [testenv]
 deps=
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -55,7 +55,7 @@ commands=
     rm -rf ./package
 
 [testenv:integration-tests]
-envlist = py37
+envlist = py39
```
rohanpm, line 58: Hmm, the problem isn't introduced by this commit, but ... I kinda doubt that this does anything useful.

I assume that the original author expected that putting `envlist = py37` in here would mean "use python 3.7 when running tox -e integration-tests".

But it doesn't mean that.  The option is documented at https://tox.wiki/en/latest/config.html#conf-envlist , and from those docs I would expect this line in the config to have no effect.  This can be confirmed locally by trying "tox -e integration-tests" and seeing that it uses whatever's the system python, not 3.7 (or 3.9).

Could we please remove it rather than leaving the misleading config in place?

rohanpm (approved):

PR #308, opened 2021-12-15
Remove listing data, references from integration tests

negillett:
Listing files are no longer expected on CDN, as listings are now generated
upon request from exodus-config data on $project-config-$env tables.

### DIFF
--- a/support/reftest/data.yml
+++ b/support/reftest/data.yml
@@ -28,10 +28,6 @@ test_data:
 - path: /origin/rpms/bash/4.4.19/8.el8_0/fd431d51/bash-4.4.19-8.el8_0.x86_64.rpm
   sha256: 4164ff2c0116d666578ba5e456ab03b88788dfb42fb93fc91b2c1da709d86686
   content-type: application/x-rpm
-# listing
-- path: /content/dist/rhel/server/5/5.7/listing
-  sha256: b51f4ddc06fddec9e73892671f1f25300ccd4235fa798afd01caf96446dc2bf1
-  content-type: text/plain
 # PULP_MANIFEST
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/PULP_MANIFEST
   sha256: 2fb7a09fd67432ad7b7c95d24fe332827e24920347319c24bfef236cb0b72f19
--- a/tests/integration/test_exodus.py
+++ b/tests/integration/test_exodus.py
@@ -30,7 +30,6 @@ def test_header_not_exist_file(cdn_test_url):
 # Test data for exodus-lambda
 # serving repo entry points with appropriate cache headers
 testdata_cache_control_path = [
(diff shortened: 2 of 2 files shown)


negillett:
This fixes the currently broken pipeline tests.

We may want to expand integration tests to also cover releasever and listing generation which would involve initializing the 'config-table' with some data like that in `./test/test_data/exodus-config.json`.
@dichen16, wdyt about adding that to https://issues.redhat.com/browse/RHELDST-7991?

tests/integration/test_exodus.py
```diff
@@ -30,7 +30,6 @@ def test_header_not_exist_file(cdn_test_url):
 # Test data for exodus-lambda
 # serving repo entry points with appropriate cache headers
 testdata_cache_control_path = [
-    "/content/dist/rhel/server/5/5.7/listing",
```
rohanpm, line 33: I don't think this should be removed.  I think a test failure here is accurately showing that we missed something:

If you have a look in the origin response function, you'll see that listing files are some of those where we intentionally set the cache-control header.  However since we now generate the listing responses in origin request, it means that origin response is no longer reached for listings.  So we've lost the cache-control header for listings.

The test has detected this bug, I think the appropriate response is to update the listing code in origin request to restore the lost header, and not silence the test.

support/reftest/data.yml
```diff
@@ -28,10 +28,6 @@ test_data:
 - path: /origin/rpms/bash/4.4.19/8.el8_0/fd431d51/bash-4.4.19-8.el8_0.x86_64.rpm
   sha256: 4164ff2c0116d666578ba5e456ab03b88788dfb42fb93fc91b2c1da709d86686
   content-type: application/x-rpm
-# listing
```
rohanpm, line 31: I think we should not remove this but instead prevent it from being uploaded.

I see the reftest/data.yml as serving two purposes:

1. Declares the set of data which can be copied from legacy CDN to populate exodus CDN for testing.

2. Declares a set of endpoints which, given a healthy CDN, should be possible to query and observe the specified properties.

Since we don't upload listing files by traditional means any more then indeed listing isn't valid for (1) now.  But it's still valid for (2), isn't it?  i.e. even with the new method of handling listings, it's still expected that I should be able to query /content/dist/rhel/server/5/5.7/listing and get some content of sha256sum b51f4ddc06fddec9e73892671f1f25300ccd4235fa798afd01caf96446dc2bf1 ?

In that case I don't think we should purge it from data, because it's a valuable test case.  (I thought we already had a test of the form "iterate through all these entries, query the path, verify we got a 200 response, verify checksum if defined" - and if we don't have that we probably should.)

Could we instead do something like `deploy: false` and then update support/reftest/reftest to skip any item where deploy is false?  So it could continue to serve both the above use-cases.

rohanpm (changes requested):

PR #316, opened 2022-01-17
Decouple integration tests with lambda conf

dichen16:


negillett (approved):

tests/integration/test_exodus.py
```diff
@@ -42,7 +42,7 @@ def test_header_cache_control(cdn_test_url, testdata_path):
     r = requests.get(url)
     print(json.dumps(dict(r.headers), indent=2))
     assert r.status_code == 200
-    assert r.headers["cache-control"] == "max-age=600"
+    assert r.headers["cache-control"].startswith("max-age=")
```
rohanpm, line 45: What do you think of doing a pattern match on the value?  e.g. verify that it matches regex `^max-age=[0-9]+$`.

The concern I have with the current approach is that it makes the test a bit too weak, e.g. if we completely failed to substitute in the config, or we had a formatting issue like `max-age="600"` rather than `max-age=600`, the test would still pass.

dichen16:
Passed AWS CodePipeline of my fork: https://console.aws.amazon.com/codesuite/codepipeline/pipelines/dichen-pipeline-dev/view?region=us-east-1

rohanpm (approved):

PR #320, opened 2022-02-14
Return 404 responses for 'absent' items [RHELDST-9952]

dichn:
In order to support the explicit deletion/hiding of content on
exodus-cdn, it should make exodus-lambda return 404 responses for URIs
which have been marked as absent.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -168,15 +168,19 @@ class OriginRequest(LambdaBase):
             self.logger.info("Item found for '%s'", uri)
 
             try:
+                # Validate If the item's "object_key" is "absent"
+                object_key = query_result["Items"][0]["object_key"]["S"]
+                if object_key == "absent":
+                    self.logger.info("Item absent for '%s'", uri)
+                    return {"status": "404", "statusDescription": "Not Found"}
+
                 # Add custom header containing the original request uri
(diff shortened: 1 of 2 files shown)


negillett (dismissed):

crungehottman (commented):
I think this patch is missing an integration test for an item with object_key == 'absent'

exodus_lambda/functions/origin_request.py
```diff
@@ -165,18 +165,22 @@ def handler(self, event, context):
...
         if query_result["Items"]:
-            self.logger.info("Item found for '%s'", uri)
-
             try:
+                # Validate If the item's "object_key" is "absent"
+                object_key = query_result["Items"][0]["object_key"]["S"]
+                if object_key == "absent":
+                    self.logger.info("Item absent for '%s'", uri)
+                    return {"status": "404", "statusDescription": "Not Found"}
+                else:
```
rohanpm, line 174: Can the "else" please be dropped to reduce unnecessary nesting?

It has no effect since the "if" case always returns.  It's also arbitrary that you put only one line of code in it.  Every line of code between here and "return request" should execute only in the "item found" case, what's special about this one line of code which makes sense for it to be in "else" and not any of the others?

rohanpm (approved):

negillett (approved):

PR #321, opened 2022-02-15
reftest: support absent items

dichn:


crungehottman (dismissed):

negillett (approved):

support/reftest/data.yml
```diff
@@ -41,3 +41,5 @@ test_data:
 - path: /content/dist/rhel/atomic/7/7Server/x86_64/ostree/repo/refs/heads/rhel-atomic-host/7/x86_64/standard
   content-type: text/plain
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/rhel-8.2-x86_64-boot.iso
+- path: /content/dist/rhel8/8.2/x86_64/baseos/iso/does-not-exist.iso
+  absent: true
```
rohanpm, line 45: While not a big deal, I think the way of representing an absent item could have been better.  I believe either of the following options would have been better:

- Use 'state: absent' instead.  This would provide a nice consistency between this and ansible, as almost all ansible modules which can make something present or absent will use a 'state' parameter which defaults to 'present'.
- Or use 'sha256: absent' instead.  This would fix the problem that it's currently possible to mix both "sha256: (something)" and "absent: true" together and it's not obvious what that's supposed to do.  (If the sha256 I specify in that case will be simply thrown away, then why am I allowed to provide it in the first place?)

rohanpm (dismissed):

rohanpm (approved):

negillett (approved):

PR #323, opened 2022-03-01
Use signed cookies in integration tests [RHELDST-8860]

dichn:
If tests are run without the two env vars "EXODUS_CDN_PRIVATE_KEY"
and "EXODUS_CDN_PRIVATE_KEY", requests will not be signed.

### DIFF
--- a/tests/test_utils/utils.py
+++ b/tests/test_utils/utils.py
@@ -1,5 +1,8 @@
 import json
 import os
+from datetime import datetime, timedelta, timezone
+
+from exodus_lambda.functions.signer import Signer
 
 CONF_FILE = os.environ.get("EXODUS_LAMBDA_CONF_FILE")
 
@@ -50,3 +53,22 @@ def mock_definitions():
         exodus_config = json.load(f)
 
     return exodus_config
+
+
+def generate_test_cookies():
+    # Env var for using signed cookies
+    key = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
+    key_id = os.environ.get("EXODUS_CDN_KEY_ID")
+
+    if not key or not key_id:
+        # envvar absent, requests will not be signed
+        return {}
+
+    expiration = datetime.now(timezone.utc) + timedelta(hours=1)
+    signer = Signer(key, key_id)
+    cookies = signer.cookies_for_policy(
+        append="",
+        resource="https://*",
+        date_less_than=expiration,
+    )
+    return {item.split("=")[0]: item.split("=")[1] for item in cookies}
--- a/tests/integration/test_exodus.py
+++ b/tests/integration/test_exodus.py
@@ -4,6 +4,10 @@ import re
 import pytest
 import requests
 
+from ..test_utils.utils import generate_test_cookies
+
+TEST_COOKIES = generate_test_cookies()
+
 
 def test_exodus_basic(cdn_test_url):
     url = (
@@ -11,7 +15,7 @@ def test_exodus_basic(cdn_test_url):
         + "/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm"
     )
 
-    r = requests.get(url)
+    r = requests.get(url, cookies=TEST_COOKIES)
     print(json.dumps(dict(r.headers), indent=2))
     assert r.status_code == 200
     assert "cache-control" not in r.headers
@@ -22,7 +26,7 @@ def test_header_not_exist_file(cdn_test_url):
         cdn_test_url
         + "/content/aus/rhel/server/6/6.5/x86_64/os/Packages/c/cpio-2.10-12.el6_5.x86_64.rpm_not_exist"  # noqa: E501
     )
-    r = requests.get(url)
+    r = requests.get(url, cookies=TEST_COOKIES)
     print(json.dumps(dict(r.headers), indent=2))
     assert r.status_code == 404
     assert "cache-control" not in r.headers
@@ -40,7 +44,7 @@ testdata_cache_control_path = [
 @pytest.mark.parametrize("testdata_path", testdata_cache_control_path)
 def test_header_cache_control(cdn_test_url, testdata_path):
     url = cdn_test_url + testdata_path
-    r = requests.get(url)
+    r = requests.get(url, cookies=TEST_COOKIES)
     print(json.dumps(dict(r.headers), indent=2))
     assert r.status_code == 200
     assert re.match("^max-age=[0-9]+$", r.headers["cache-control"])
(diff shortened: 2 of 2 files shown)


dichn:
@negillett @crungehottman @rohanpm 
I think this PR can only be merged until the corresponding ansible playbook supports deploying signing keys. Then, the integration tests can pass (I tested it on my CloudFront distribution which configured the signing keys manually in AWS web console).
This patch handles:
(1) Read the private-key and key-pair-id from env var
(2) Generate signed cookies
(3) Send requests to CloudFront with the signed cookies

negillett:
> @negillett @crungehottman @rohanpm I think this PR can only be merged until the corresponding ansible playbook supports deploying signing keys. Then, the integration tests can pass (I tested it on my CloudFront distribution which configured the signing keys manually in AWS web console). This patch handles: (1) Read the private-key and key-pair-id from env var (2) Generate signed cookies (3) Send requests to CloudFront with the signed cookies

Yeah, I think a lot of this depends on RHELDST-8859. And I nearly have it ready for code review.

tests/test_utils/utils.py
```diff
@@ -52,3 +60,50 @@ def mock_definitions():
...
+    loaded_key = serialization.load_pem_private_key(
+        bytes_key, password=None, backend=default_backend()
+    )
+    return loaded_key.sign(policy, padding.PKCS1v15(), hashes.SHA1())  # type: ignore
+
+
+def generate_cookies():
+    # Env var for using signed cookies
+    CDN_PRIVATE_KEY = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
+    CDN_KEY_ID = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
```
negillett, line 94: Should be looking for EXODUS_KEY_ID here

tests/test_utils/utils.py
```diff
@@ -52,3 +60,50 @@ def mock_definitions():
...
+    bytes_key = bytes(private_key, "utf-8")
+    loaded_key = serialization.load_pem_private_key(
+        bytes_key, password=None, backend=default_backend()
+    )
+    return loaded_key.sign(policy, padding.PKCS1v15(), hashes.SHA1())  # type: ignore
+
+
+def generate_cookies():
+    # Env var for using signed cookies
+    CDN_PRIVATE_KEY = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
```
negillett, line 93: The cookie key will be retrieved from AWS secret manager, FYI. It should be safe to use the same method for integration testing within CodePipeline.

tests/test_utils/utils.py
```diff
@@ -52,3 +60,50 @@ def mock_definitions():
...
+
+def rsa_signer(private_key: str, policy: bytes):
+    bytes_key = bytes(private_key, "utf-8")
+    loaded_key = serialization.load_pem_private_key(
+        bytes_key, password=None, backend=default_backend()
+    )
+    return loaded_key.sign(policy, padding.PKCS1v15(), hashes.SHA1())  # type: ignore
+
+
+def generate_cookies():
```
negillett, line 91: I've implemented signing code in exodus_lambda/functions/signer.py in #324

negillett: That PR may enable deployment testing key(s) for the integration tests.

tests/test_utils/utils.py
rohanpm: I'd rather this just looks at env vars and doesn't directly use secrets manager, as we should ideally be able to easily run these tests locally also, maybe even against a non-AWS-hosted dev server.

It seems like, if the private key is in secrets manager, it will be an option to set up the codepipeline to take it from SM and put it in this env var - from https://aws.amazon.com/about-aws/whats-new/2019/11/aws-codebuild-adds-support-for-aws-secrets-manager/ . Haven't looked into it further than that.

rohanpm, line 93: Minor: local vars shouldn't be named in all caps.  You'd name these CDN_PRIVATE_KEY etc if they were globals, which they aren't.

rohanpm (changes requested):

tests/test_utils/utils.py
dichn: @negillett 
if using the exodus_lambda/functions/signer.py and its `signer.cookies_for_policy()`.
```
        cookies_content = signer.cookies_for_policy(
            append="; Secure; Path=/content/; Max-Age=%s"
            % int(expiration.total_seconds()),
            resource="https://%s/content/*" % domain,
```
I think it requires integration tests ask for cookies for every request (`r = requests.get(url, cookies=TEST_COOKIES)`)
The AC for integration tests is one cookies for all requests, so maybe not using the new cookies endpoint in #324, WDYT?
CC: @crungehottman @rohanpm

negillett: @dichn, 
My point is only that some of the stuff from #324 can be reused here.
I.e., you could adapt the `generate_cookies` function. Something like this should work;
```
def generate_test_cookies():
    key = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
    key_id = os.environ.get("EXODUS_CDN_KEY_ID")

    if not key_id or not key:
        return {}

    signer = Signer(key, key_id)
    expiration = datetime.utcnow() + timedelta(minutes=60)
    return signer.cookies_for_policy(date_less_than=expiration)
```

dichn:
Status:
* move `CloudFrontSigner::build_policy()` out of `Signer::cookies_for_policy`, so tests can do some customization to the policy (remove the "resource" parameter, so `TEST_COOKIES` can access all the test content)
* fixed the type in env vars reading

negillett:
> * move `CloudFrontSigner::build_policy()` out of `Signer::cookies_for_policy`, so tests can do some customization to the policy (remove the "resource" parameter, so `TEST_COOKIES` can access all the test content)

This works but seems unnecessary. Instead of doing all that work to omit/remove resource from the policy, you just needed to add one accepted by all test requests, e.g., "https://*".
Applying that to my earlier suggestion:
```
def generate_test_cookies():
    key = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
    key_id = os.environ.get("EXODUS_CDN_KEY_ID")

    if not key_id or not key:
        return {}

    signer = Signer(key, key_id)
    expiration = datetime.utcnow() + timedelta(minutes=60)
    return signer.cookies_for_policy(append="", resource="https://*", date_less_than=expiration)
```

dichn:
@negillett many thank!!! It works and much simpler than the previous change.

tests/test_utils/utils.py
```diff
@@ -50,3 +53,23 @@ def mock_definitions():
...
+def generate_test_cookies():
+    # Env var for using signed cookies
+    key = os.environ.get("EXODUS_CDN_PRIVATE_KEY")
+    key_id = os.environ.get("EXODUS_CDN_KEY_ID")
+
+    if not key or not key_id:
+        # envvar absent, requests will not be signed
+        return {}
+
+    # The cookies expire in 1 hour
```
rohanpm, line 67: Minor: it's a bit funny to write "expire in 1 hour", then "timedelta(seconds=3600)" when you could just write "timedelta(hours=1)".  The main point of that class seems to be it allows you to deal with units other than seconds if you want.

rohanpm (approved):

rohanpm (approved):

PR #324, opened 2022-03-01
Add signed cookie endpoint [RHELDST-8859]

negillett:
The new endpoint enables signed URls/cookies onto Exodus CloudFront
distributions. One can now request "/_/cookie/<url>" of the distribution
to generate a signed cookie and be redirected to the requested URL.

This commit also updates configuration and cloudformation templates to
appropriately support this feature.


configuration/exodus-lambda-deploy.yaml
```diff
@@ -36,7 +55,7 @@ Resources:
         Name: !Sub ${project}-cache-policy-${env}
         ParametersInCacheKeyAndForwardedToOrigin:
           CookiesConfig:
-            CookieBehavior: none
+            CookieBehavior: all
```
rohanpm, line 58: Can we please use a whitelist rather than "all" ?   Setting "all" may make the cache less effective as responses will not be cacheable across, e.g., two different browsers which are passing some random cookies unrelated to exodus.

configuration/lambda_config.template
```diff
@@ -11,10 +11,13 @@
       "us-east-1"
     ]
   },
-  "config_cache_ttl": $EXODUS_CONFIG_CACHE_TTL,
+  "config_cache_ttl": "$EXODUS_CONFIG_CACHE_TTL",
```
rohanpm, line 14: Was there a reason you wanted to wrap this and the next one in quotes?  The values to be substituted are integers.  It doesn't seem necessary to wrap them in strings just to unwrap them again later.

I think we've been through this before, there was confusion around this file not being valid JSON (which it is not if there's no quotes around the $ placeholders).  But that's why it was renamed to .template to more accurately reflect that it's not a JSON file, it's a template which can be processed in order to generate a JSON file.  If you had to add the quotes to make something work, there was probably a bug somewhere else.

exodus_lambda/functions/origin_request.py
```diff
@@ -103,6 +149,40 @@ def resolve_aliases(self, uri):
...
+        signer = Signer(self.cookie_key, self.conf.get("key_id"))
+        expiration = timedelta(minutes=int(self.conf.get("cookie_ttl", "720")))
+        domain = event["Records"][0]["cf"]["config"]["distributionDomainName"]
+        uri = event["Records"][0]["cf"]["request"]["uri"]
+
+        cookies_content = signer.cookies_for_policy(
+            append="; Secure; Path=/content/; Max-Age=%s"
+            % int(expiration.total_seconds()),
+            resource="https://%s/content/*" % domain,
+            date_less_than=datetime.utcnow() + expiration,
```
rohanpm, line 162: Would you mind calling utcnow() just once in here, so that the cookies always come in a pair?

If it's called more than once, time can pass between the two which means the origin and content cookies can expire at slightly different times.  Not a big deal, it just makes it harder to match them up as "these were generated together" while debugging.

exodus_lambda/functions/origin_request.py
```diff
@@ -16,14 +17,66 @@ class OriginRequest(LambdaBase):
...
+                "secretsmanager", region_name=self.region
+            )
+
+        return self._sm_client
+
+    @property
+    def secret(self):
+        if not self._cache.get("secret"):
+            arn = self.conf.get("secret_arn")
+            self.logger.warning("attempting to get secret %s", arn)
```
rohanpm, line 49: These shouldn't be warnings, they might have been set that way during development, but debug makes more sense going forward.

rohanpm (changes requested):

configuration/exodus-lambda-deploy.yaml
negillett: Yeah, I'm just not sure what should be listed. From what I can tell, our cookies don't have names. Or it looking for domain names? Documentation doesn't specify, unfortunately, what's sought.

exodus_lambda/functions/origin_request.py
negillett: Actually, it seems we've chosen to use info for everything within lambda functions as we only look through these logs in CloudWatch when debugging.

negillett: There's one debug statement and it's for an error. So, yeah, it seems logging scope is shifted for lambda.

configuration/lambda_config.template
negillett: Yeah, I think there was a bug somewhere in the testing setup but it seems to be resolved now.

configuration/exodus-lambda-deploy.yaml
rohanpm: The cookie names are those starting with "CloudFront-", e.g. "CloudFront-Key-Pair-Id", from cookies_for_policy method.

exodus_lambda/functions/origin_request.py
```diff
@@ -24,6 +26,57 @@ def __init__(self, conf_file=CONF_FILE):
...
+            #        "string",
+            #    ],
+            #    "CreatedDate": datetime(2015, 1, 1)
+            #  }
+            #
+            # We're interested in SecretString and expect it to be a JSON string
+            # containing the "cookie_key" key.
+            if "SecretString" in response:
+                secret = response["SecretString"]
+                self.logger.info("secret string %s", repr(secret)[0:50])
```
rohanpm, line 69: This one really shouldn't be info I think.  I believe I wrote this originally?  It was to help me understand what SM was returning during development, but given this is an actual secret I don't think we should permanently be logging it, even if it's just the first 50 chars.  Could you please drop it or make it debug?

negillett: Thanks for pointing this out. Improving logging all around now.

configuration/exodus-lambda-deploy.yaml
negillett: Ah, okay. Thanks!

rohanpm (approved):

PR #325, opened 2022-03-09
Fix a template error

dichn:
CF stack "exodus-lambda-dev" throw a template error: every Fn::Equals
object requires a list of 2 string parameters.

(1) Cast keyids(CommaDelimitedList) to String
(2) None -> ""

### DIFF
--- a/configuration/exodus-lambda-deploy.yaml
+++ b/configuration/exodus-lambda-deploy.yaml
@@ -32,8 +32,8 @@ Parameters:
 Conditions:
   EnableKeyGroup: !Not
     - !Equals
-      - !Ref keyids
-      - None
+      - !Join ["", !Ref keyids]
+      - ""
 
 Resources:
   KeyGroup:


negillett (approved):

configuration/exodus-lambda-deploy.yaml
```diff
@@ -32,8 +32,8 @@ Parameters:
 Conditions:
   EnableKeyGroup: !Not
     - !Equals
-      - !Ref keyids
-      - None
+      - !Join ["", !Ref keyids]
```
rohanpm, line 35: I don't object to the fix if it works, but I find this a little suspicious because here's the doc for Equals: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html#intrinsic-function-reference-conditions-equals

And it says there about the parameters:

> A value of any type that you want to compare.

It's explicitly calling out that it should work on "any type".  So something seems to be wrong here, either in the docs or in our understanding of the problem.

rohanpm (approved):

configuration/exodus-lambda-deploy.yaml
dichn: @rohanpm I think the error message `Template error: every Fn::Equals object requires a list of 2 string parameters.` is conflicted with this statement `A value of any type that you want to compare.`
So I think the doc may be wrong.

PR #326, opened 2022-03-10
Fix condition in lambda deploy template

negillett:
Previously, the EnableKeyGroup condition in exodus-lambda-deploy.yaml
checked that a joined list of keyids was an empty string. However, the
default value of keyids is the string "None", so the joined keyids
string will never be "".

The failure of this condition meant that an attempt was made to create
a KeyGroup resource even when no valid IDs existed.

This commit corrects the condition to look for the string "None".

### DIFF
--- a/configuration/exodus-lambda-deploy.yaml
+++ b/configuration/exodus-lambda-deploy.yaml
@@ -30,10 +30,8 @@ Parameters:
     Description: A list of secret key IDs for Distribution key groups
 
 Conditions:
-  EnableKeyGroup: !Not
-    - !Equals
-      - !Join ["", !Ref keyids]
-      - ""
+  EnableKeyGroup:
+    !Not [!Equals [!Join ["", !Ref keyids], None]]
 
 Resources:
   KeyGroup:


configuration/exodus-lambda-deploy.yaml
```diff
@@ -30,10 +30,8 @@ Parameters:
     Description: A list of secret key IDs for Distribution key groups
 
 Conditions:
-  EnableKeyGroup: !Not
-    - !Equals
-      - !Join ["", !Ref keyids]
-      - ""
+  EnableKeyGroup:
+    !Not [!Equals [!Join ["", !Ref keyids], None]]
```
rohanpm, line 34: This probably works but maybe the default value should be an empty list?  Not sure how that looks for CommaDelimitedList, perhaps `""`.  For a list of key IDs, "empty list" seems a better default value than either a string "None" or an actual `None`/`null`.

rohanpm (approved):

configuration/exodus-lambda-deploy.yaml
negillett: Yeah, any string will work but None is already used in the template for default values so I stuck with that.

It doesn't really make sense as a default for any of the other parameters either. Maybe we should follow up with a PR to update them all.

dichn:
I was actually preferring the following format. But I don't object to this fix if it works.
```
Conditions:
  EnableKeyGroup: !Not
    - !Equals
      - !Join ["", !Ref keyids]
      - None
```

dichn (approved):

PR #327, opened 2022-03-18
Fix distribution deployment

negillett:
Cache behavior of distributions require a list of trusted key groups,
but was being given a string value in our template,
exodus-lambda-deploy.yaml. This commit corrects the mistake.

### DIFF
--- a/configuration/exodus-lambda-deploy.yaml
+++ b/configuration/exodus-lambda-deploy.yaml
@@ -86,7 +86,7 @@ Resources:
               LambdaFunctionARN: !Ref OriginResponseFunc.Version
           TargetOriginId: !Sub S3-${project}-cdn-${env}
           TrustedKeyGroups:
-            !If
+            - !If
               - EnableKeyGroup
               - !Ref KeyGroup
               - !Ref AWS::NoValue


rohanpm (approved):

configuration/exodus-lambda-deploy.yaml
```diff
@@ -86,7 +86,7 @@ Resources:
               LambdaFunctionARN: !Ref OriginResponseFunc.Version
           TargetOriginId: !Sub S3-${project}-cdn-${env}
           TrustedKeyGroups:
-            !If
+            - !If
```
rohanpm, line 89: I had approved this, but thinking on it some more I have doubts whether the NoValue case works now.

I find the types in use easier to understand in JSON format rather than YAML.  I think it was like this prior to your change:

```json
  "TrustedKeyGroups": {
    "Fn::If": [
      "EnableKeyGroup",
      {"Ref": "KeyGroup"},
      {"Ref": "AWS::NoValue"},
    ]
  }
```

What happens then is, when functions are evaluated the entire Fn::If object is replaced with one of the two possible cases, so it's going to be either:

```json
// case 1
"TrustedKeyGroups": {"Ref": "KeyGroup"},

// or, case 2
"TrustedKeyGroups": {"Ref": "AWS::NoValue"},
```

Case 1 was definitely wrong as described in the commit message, as TrustedKeyGroups is supposed to take a list and "Ref KeyGroup" is going to give us a string, not a list.  But I think that case 2 was not wrong.

Now with your change, it's going to be like this:

```json
  "TrustedKeyGroups": [ {
    "Fn::If": [
      "EnableKeyGroup",
      {"Ref": "KeyGroup"},
      {"Ref": "AWS::NoValue"},
    ]
  } ]
```
After evaluation of the If:

```json
// case 1
"TrustedKeyGroups": [{"Ref": "KeyGroup"}],

// or, case 2
"TrustedKeyGroups": [{"Ref": "AWS::NoValue"}],
```

Now case 1 has been fixed - but was case 2 broken?  I think it might be.  Docs for NoValue are at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-novalue and it doesn't say anything about what happens if you put NoValue into a list.  My guess would be that it's not valid to do so, and NoValue always has to be used at a resource property level.

I think we want the If to return a list in the enabled case, and NoValue in the disabled case, which would look like this in the template:

```json
  "TrustedKeyGroups": {
    "Fn::If": [
      "EnableKeyGroup",
      [{"Ref": "KeyGroup"}],
      {"Ref": "AWS::NoValue"},
    ]
  }
```

Or, in YAML:

```yaml
  TrustedKeyGroups:
    !If
      - EnableKeyGroup
      - - !Ref KeyGroup
      - !Ref AWS::NoValue
```

Could you please test the "key group not enabled" case if you haven't already and let me know if it works?

rohanpm (changes requested):

configuration/exodus-lambda-deploy.yaml
negillett: Nice catch! Yeah, I hadn't noticed this and it seems like it would be an issue but deployment succeeds regardless. It seems CloudFormation is smart enough to scrap TrustedKeyGroups when it reads NoValue.

negillett: To test this I just removed '!Not' from the EnableKeyGroup condition. This caused the deployed Distribution to be deployed with "No" under "Restrict viewer access", as should be expected.

negillett:
BTW, pipelines are still failing due to missing SecretsManager permission on CodeBuild role. I've filed a ticket with IT to have it and the Lambda role updated with GetSecret access.

rohanpm (approved):

PR #338, opened 2022-04-07
Add "X-Exodus-Query" header

dichn:
Because of the traffic shift and traffic will be routed to different
lambda versions (in different regions), integration tests may run again
an old version lambda.

The "X-Exodus-Query" header is primarily used in a request, to provide
the version of exodus-lambda, "X-Exodus-Version" in response.

For traffic shift:
https://aws.amazon.com/blogs/compute/implementing-safe-aws-lambda-deployments-with-aws-codedeploy/

### DIFF
--- a/exodus_lambda/functions/origin_response.py
+++ b/exodus_lambda/functions/origin_response.py
@@ -33,6 +33,11 @@ class OriginResponse(LambdaBase):
                 {"key": "Digest", "value": f"id-sha-256={sum_b64}"}
             ]
 
+        if "headers" in request and "x-exodus-query" in request["headers"]:
+            response["headers"]["x-exodus-version"] = [
+                {"key": "X-Exodus-Version", "value": self.lambda_version}
+            ]
+
         try:
             original_uri = request["headers"]["exodus-original-uri"][0][
                 "value"
--- a/configuration/exodus-lambda-deploy.yaml
+++ b/configuration/exodus-lambda-deploy.yaml
@@ -63,6 +63,7 @@ Resources:
             HeaderBehavior: whitelist
             Headers:
               - Want-Digest
+              - X-Exodus-Query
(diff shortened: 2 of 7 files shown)


dichn:
re-push for fixing a typo in commit message (no code change)

exodus_lambda/functions/origin_response.py
```diff
@@ -33,6 +33,12 @@ def handler(self, event, context):
                 {"key": "Digest", "value": f"id-sha-256={sum_b64}"}
             ]
 
+        if "headers" in request and "x-exodus-query" in request["headers"]:
+            self.logger.error("x-exodus-query exist")
```
rohanpm, line 37: Please drop this as it's not an error.  You could keep a debug() if you really need it.

rohanpm (changes requested):
I think the commit message is misleading:

- this has nothing to do with
https://aws.amazon.com/blogs/compute/implementing-safe-aws-lambda-deployments-with-aws-codedeploy/ , which is not something we're using or will be able to use
- I don't think "Because of the traffic shift and traffic will be routed to different
lambda versions (in different regions)" is correct.  At best it's misleading to say this is because of "traffic shift" and then link to an article about a kind of traffic shifting we're not doing.

Could the commit message please stick with the known facts?  You don't have to put forward a theory as to why we might be hitting an old version of the lambda sometimes.

I would have written something like "Our tests sometimes seem to hit an old version of our lambda, for unclear reasons. Let's make it possible to query the currently deployed lambda version so we can at least reliably detect when this happens."

dichn:
1) commit message updated
2) remove useless log

rohanpm (dismissed):

dichn:
re-push after the rebase of #337

crungehottman (approved):

PR #347, opened 2022-04-12
Integration tests predicate on source version check

dichn:


tests/conftest.py
```diff
@@ -47,6 +53,17 @@ def cdn_test_url(request):
     return url
 
 
+@pytest.fixture
+def src_version(request):
```
rohanpm, line 57: Minor: this could do with a comment or doc string.  The meaning is not obvious.

tests/conftest.py
```diff
@@ -47,6 +53,17 @@ def cdn_test_url(request):
...
+    if request.config.getoption("--src-version"):
+        version = request.config.getoption("--src-version")
+    elif os.environ.get("CODEBUILD_RESOLVED_SOURCE_VERSION"):
+        version = os.environ.get("CODEBUILD_RESOLVED_SOURCE_VERSION")
+    else:
+        pytest.skip("Test skipped because failed to get lambda version")
```
rohanpm, line 63: It still shouldn't be skipping if the version is unknown.

tests/conftest.py
```diff
@@ -105,3 +122,31 @@ def dummy_private_key():
...
+                "Expected to run against version %s but server sent X-Exodus-Version: %s"
+                % (self.expected_version, version)
+            )
+        return response
+
+
+@pytest.fixture
+def requests_session(src_version):
+    session = requests.Session()
+    if src_version:
```
rohanpm, line 148: This seems like dead code, because you didn't write src_version to return None if there's no version.  You wrote it to call pytest.skip, so you'd never get here.

rohanpm (changes requested):

dichn:
update:
1. add comments for fixture: src_version
2. when src_version is unknown, the version check will be skipped

rohanpm (approved):

PR #348, opened 2022-04-12
refactor: ensure X-Exodus-Version works for synthesized responses too

rohanpm:
If the logic for setting X-Exodus-Version lives only in origin-response,
it can't work for responses generated directly without going to origin,
e.g. /listing files, /_/cookie, 404 errors and maybe some other things
in the future.

Implement it also in origin-request, in a way which ensures it can't be
forgotten as handler is extended.


rohanpm:
I'm closing this draft as I only opened it to share an implementation approach here, it wasn't intended for merge.  Please cherry-pick it or similar into another PR if desired.

PR #355, opened 2022-04-20
Revise the env var scheme for log levels [RHELDST-11210]

rohanpm:
During testing I learned that the previous scheme for allowing log
levels to be set by env vars would not work in practice, due the
following restriction when setting Lambda environment variables:

```
  Member must satisfy regular expression pattern: [a-zA-Z]([a-zA-Z0-9_])+]
```

The restriction meant that loggers with a name containing characters like '-'
or '.' could not be configured.

Revise the scheme so that the logger name does not need to appear in the
environment variable name.


rohanpm:
This was merged accidentally and then backed out.  It is not actually present in master now.

PR #381, opened 2022-08-01
Look up and serve index pages automatically [RHELDST-12446]

crungehottman:


exodus_lambda/functions/origin_request.py
```diff
@@ -19,6 +19,8 @@
 # this code against localstack.
 ENDPOINT_URL = os.environ.get("EXODUS_AWS_ENDPOINT_URL") or None
 
+INDEX_FILENAME = "__exodus_autoindex__"
```
negillett, line 22: Elsewhere you have the name as ".__exodus_autoindex". Should it be the same here?

crungehottman: forgot to remove this, line - ty

exodus_lambda/functions/origin_request.py
```diff
@@ -290,52 +267,85 @@ def handler(self, event, context):
...
+        )
+        self.logger.info(
+            "The uri value for origin_request beginning is '%s'", uri
+        )
+        table = self.conf["table"]["name"]
 
-            # Report 404 to prevent attempts on S3
-            return {"status": "404", "statusDescription": "Not Found"}
+        # Do not permit clients to explicitly request an index file
+        if not uri.endswith(self.index):
```
rohanpm, line 339: I think this should be endswith('/' + self.index), right?  Otherwise it's blocking the fetch of all files which end in the configured filename, while the desired behavior is to block just the fetch of files whose name precisely matches the index.

rohanpm (changes requested):

negillett (approved):

rohanpm (approved):

PR #389, opened 2022-08-25
Ensure URL always ends with / when serving index [RHELDST-13170]

rohanpm:
When trying to emulate directories by serving up index pages for certain
URIs it is important that the browser has requested the URL of the 
"directory" while including a trailing /. If not, any relative links
in an HTML index will not resolve as desired.
We therefore need to send a redirect to the requested URL with /
appended if it wasn't already present.

As part of this fix, the autoindex test also had to be fixed as it was 
not actually covering autoindex. It had mocked the DynamoDB query to
always return a result, but in reality the autoindex scenario first does
one query which fails to find a result, and then a second query to look
up the autoindex.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -352,6 +352,34 @@ class OriginRequest(LambdaBase):
             index_uri = index_uri + "/" + self.index
             for query_uri in (uri, index_uri):
                 if out := self.response_from_db(request, table, query_uri):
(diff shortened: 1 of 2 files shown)


rohanpm:
This PR was originally just about fixing the issue with the test, I changed it to cover RHELDST-13170 fully after realizing there would be git conflicts otherwise.

crungehottman (approved):

negillett (approved):

PR #405, opened 2022-11-11
Integrate bandit SAST tool with CI [RHELDST-14783]

dichn:


.github/workflows/ci.yml
```diff
@@ -22,7 +22,10 @@ jobs:
       run: |
         sudo apt install graphviz
         python -m pip install -U pip
-        pip install tox
+        pip install tox bandit
```
rohanpm, line 25: This should really be installed inside of tox.  That way it can be installed automatically when people run 'tox' locally, and also the version of it can be pinned and managed same as our other dependencies.

Can you please do it that way?  Add it to test-requirements.in and test-requirements.txt, same as other testing tools like mypy, pytest, pylint.

rohanpm (changes requested):

.github/workflows/ci.yml
rohanpm: Also, this is what the following messages in the log are referring to.

```
bandit-test run-test: commands[0] | bandit -r -ll -ii .
WARNING: test command found but not installed in testenv
  cmd: /opt/hostedtoolcache/Python/3.9.15/x64/bin/bandit
  env: /home/runner/work/exodus-lambda/exodus-lambda/.tox/bandit-test
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```

tox.ini
```diff
@@ -25,6 +25,11 @@ usedevelop=true
 commands=
 	pytest --cov-fail-under=100 --cov=exodus_lambda {posargs}
 
+[testenv:bandit-test]
+usedevelop=true
+commands=
+	bandit -r -ll -ii .
```
rohanpm, line 31: You can exclude .tox directory.

rohanpm: Also, is there a specific reason to add `-ii` ?

bandit.yaml
```diff
@@ -0,0 +1,2 @@
+skips: ['B303']
```
rohanpm, line 1: That's going to disable the check throughout the entire code, isn't it?  I don't think that's appropriate.  Even if all current violations of the rule are not a problem, what if someone introduced new code which violates this rule and it really is a problem?  It wouldn't be detected.

I think it needs to be handled instead as described in: https://bandit.readthedocs.io/en/latest/config.html#suppressing-individual-lines

rohanpm (approved):

PR #421, opened 2023-01-24
Validate request URI and query string [RHELDST-14939]

negillett:


tests/functions/test_origin_request.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import urllib
 
+import jsonschema
```
crungehottman, line 5: I think this can be removed

negillett: Thanks!

exodus_lambda/functions/origin_request.py
```diff
@@ -122,7 +122,7 @@ def secret(self):
                 self._cache["secret"] = out
                 self.logger.info("Loaded and cached secret %s", arn)
             except Exception as exc_info:
-                self.logger.error(
+                self.logger.exception(
                     "Couldn't load secret %s", arn, exc_info=exc_info
                 )
                 raise exc_info
```
rohanpm, line 128: Minor: the main point of logger.exception() is that it automatically passes exc_info of the current exception to the logger.  So this code is more complicated than it needs to be.

I believe it will work, but it would be more idiomatically written as:

```
except Exception:
    self.logger.exception("Couldn't load secret %s", arn)
    raise
```

No need to catch the exception into a variable and no need to pass exc_info argument.

exodus_lambda/functions/origin_request.py
```diff
@@ -320,10 +320,28 @@ def response_from_db(self, request, table, uri):
...
+                    request[key],
+                    exc_info=exc_info,
+                )
+                raise exc_info
+
     def handler(self, event, context):
         # pylint: disable=unused-argument
 
         request = event["Records"][0]["cf"]["request"]
+        self.validate_request(request)
```
rohanpm, line 344: I think just raising an exception here is going to give us a 500 error.  It'll be flagged as a lambda execution error.  This is not good for us because in theory we'd want to have all of those alerting us, so we can investigate our bugs.

Can we make it so that a validation failure instead results in a 400 (bad request) response code?

rohanpm (dismissed):

rohanpm:
> Also, I don't know what Bandit is complaining about. No hashes were changed. This PR is based off of upstream master. I tried running pip-compile as well to update the file and no changes resulted.

Looking at https://pypi.org/project/wcwidth/0.2.6/#files, it could be because that release originally included only a source distribution, but 2 days later added a wheel.  This effectively "changes" the hash if pip-compile originally saw the source distribution but is now trying to use the wheel.

But if just running pip-compile doesn't fix it, then there might be more than one problem.  I'll look into it and open a separate PR.

rohanpm:
> But if just running pip-compile doesn't fix it, then there might be more than one problem. I'll look into it and open a separate PR.

Hmm, I ran our usual "pip-compile" github action to create a new PR, and merged it: https://github.com/release-engineering/exodus-lambda/pull/422

Strangely it passed CI though it didn't make any changes to the hash of wcwidth package.  I don't have an explanation for that.

Anyway, given that CI was able to pass there, maybe it's worth rebasing this pull request?

exodus_lambda/functions/origin_request.py
negillett: Thanks. I'll clean up the other logger.exception calls as well.

negillett:
> Details

Yeah, rebasing worked. 🤷🏻

crungehottman (approved):

emanuelelevo (approved):

PR #429, opened 2023-02-08
Add missing Set-Cookie attributes for ESS compliance [RHELDST-14938]

negillett:
ESS guideline SEC-APP-REQ-1 requires cookies be set with a number of attributes. This commit adds the missing attributes, HttpOnly and Domain. SameSite is omitted, left to the default of 'lax', as the cookie is set by lambda but used by cloudfront. Domain is set to the distribution named in the event request's config.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -186,15 +186,16 @@ class OriginRequest(LambdaBase):
         self.logger.info("Handling cookie request: %s", uri)
 
         signer = Signer(self.cookie_key, self.conf.get("key_id"))
+
         cookies_content = signer.cookies_for_policy(
-            append="; Secure; Path=/content/; Max-Age=%s"
-            % int(ttl.total_seconds()),
+            append="; Secure; HttpOnly; SameSite=lax; Domain=%s; Path=/content/; Max-Age=%s"
+            % (domain, int(ttl.total_seconds())),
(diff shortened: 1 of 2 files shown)


negillett:
Looks like #426 is needed for this, and other PRs, to pass the pre-commit check.

rohanpm (commented):
It should be possible to get a passing CI by rebasing this now.

What's here looks right, but could you clarify about the SameSite?  If we put SameSite=Lax, would that be expected to break anything?

The issue is that, according to the docs I read at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite , it's only "recent" browsers which default the value to Lax.  The compatibility table there notably claims that Safari still doesn't have this default.

So I'm not sure we will be able to argue that we rely on the browser defaults in order to be compliant.  If SameSite=Lax works, we should probably explicitly set that (the doc above also suggests this).

negillett:
@rohanpm 
> The issue is that, according to the docs I read at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite , it's only "recent" browsers which default the value to Lax. The compatibility table there notably claims that Safari still doesn't have this default.

Ah, I missed that. Though I did see it was a recent change. Yes, we can explicitly set SameSite=lax.

rohanpm (approved):

PR #432, opened 2023-02-15
Log JSON messages by default [RHELDST-15699]

negillett:
For compatibility with splunk, exodus-lambda should produce JSON-formatted log messages. This commit alters the default formatting while leaving the ability to overwrite via environment variable.

Additionally, some log messages were shifted down from "info" level to "debug".


exodus_lambda/functions/base.py
```diff
@@ -39,6 +39,15 @@ def logger(self):
             root_logger = logging.getLogger()
             if log_conf and root_logger.handlers:
                 formatter_str = log_conf["formatters"]["default"]["format"]
+                if not formatter_str:
+                    formatter_str = json.dumps(
```
rohanpm, line 43: Hmm, never thought of doing it like this.  I think this is not safe though, because the log formatter doesn't know that JSON is being used and it won't quote values appropriately.

I mean for example, given this part:

```
"message": "%(message)s",
```

... if the message was, for example, `hello "world"` then I think we'll end up with a log record formatted as

```
{"time": ..., "name": ..., "level": ..., "message": "hello "world""}
```

...which will not be valid JSON.

I think some of the log messages are very likely to hit this problem, e.g. the "Updated request value for origin_request" message in this commit dumps an entire JSON structure in the message.  So it'll be something like:

```
"message": "Updated request value for origin_request: '{"key1": "val1", "key2": "val2", ...}'"
```

...which is not valid JSON.

I think that this has to be handled in another way, you probably cannot get a correct implementation by just setting the format string on the default formatter.  I think either the formatter or the handler has to "know" that it is generating JSON, i.e. the json.dumps should be happening within one of those objects.

rohanpm (changes requested):

PR #434, opened 2023-02-20
Convert logging format to JSON [RHELDST-15699]

negillett:
For compatibility with splunk, exodus-lambda must output machine-readable logs. This commit adds and attaches a JSON log handler with fields for log level, logger name, timestamp, and message.

Additionally, several log messages were updated for readability and some log levels were adjusted.

### DIFF
--- a/exodus_lambda/functions/base.py
+++ b/exodus_lambda/functions/base.py
@@ -4,6 +4,8 @@ import logging.config
 import os
 import re
 
+from .json_logging import JsonFormatter
+
 
 class LambdaBase(object):
     def __init__(self, logger_name="default", conf_file="lambda_config.json"):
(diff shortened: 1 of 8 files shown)


negillett:
@crungehottman @emanuelelevo @rohanpm @dichn this is now ready for review, had to toss in a small test for coverage of the new JsonHandler.

exodus_lambda/functions/base.py
```diff
@@ -38,10 +40,17 @@ def logger(self):
             logging.config.dictConfig(log_conf)
             root_logger = logging.getLogger()
             if log_conf and root_logger.handlers:
-                formatter_str = log_conf["formatters"]["default"]["format"]
-                formatter_date = log_conf["formatters"]["default"]["datefmt"]
-                formatter = logging.Formatter(formatter_str, formatter_date)
-                root_logger.handlers[0].setFormatter(formatter)
+                root_logger.handlers[0].setFormatter(
+                    JsonFormatter(
+                        fmt={
```
rohanpm, line 45: The format is not configurable?

I think we've lost aws_request_id now.  That's important, as I understand it that's the only way we'll be able to group all the log messages coming from a single lambda invocation (and hopefully trace it back to an HTTP request).

We should make sure to keep capturing the aws_request_id when present.  I can think of two ways to do that:

- make the format configurable and ensure we configure it with aws_request_id in production, or...
- (easier) include aws_request_id in the hardcoded format here, and later in formatMessage (around where `record.__dict__[v]` is accessed) make it evaluate to `None` rather than crashing when aws_request_id is absent. (So in the JSON, when running in fakefront for instance, it'll have `"aws_request_id": null`, and when running in AWS it'll have something non-null.)

rohanpm (changes requested):

exodus_lambda/functions/base.py
negillett: I've gone with the former fix and ensured normal string formatting is still supported. This way there's time to update the deployed config.

rohanpm (approved):

emanuelelevo (approved):

PR #438, opened 2023-02-28
Convert logging format to JSON [RHELDST-15699]

negillett:
For compatibility with splunk, exodus-lambda must output machine-readable logs. This commit adds and attaches a JSON log handler with fields for log level, timestamp, AWS request ID, and message.

Log message format was removed from configuration and is no longer customizable.

Additionally, several log messages were updated for readability and some log levels were adjusted.


negillett:
I didn't want to have to remove format from configuration but it seems necessary, as the file generated using a JSON env var ended up invalid and broke QA exodus-lambda.

The good news is that this should resolve the c3i logging ValueError, https://issues.redhat.com/browse/RHELDST-15869.

configuration/lambda_config.template
```diff
@@ -26,7 +26,6 @@
     "disable_existing_loggers": "False",
     "formatters": {
       "default": {
-        "format": "$EXODUS_LOG_FORMAT",
```
rohanpm, line 29: The other solution would be to make sure the format string is quoted appropriately.  Really, the problem we hit is not specific to usage of JSON - attempting to use any log format containing a `"` character or various other characters would have failed for the same reason.

That being said, I like the simplicity of the current solution.

rohanpm (approved):

configuration/lambda_config.template
negillett: Yeah, it'sas tough with bash and JSON's different issues surrounding single and double quotes.
`{'item': 'value'}` may have worked in bash, for instance, but JSON won't accept it.

I think that the need for JSON formatted logs for splunk helps further justify removing log format configuration too.

crungehottman (approved):

emanuelelevo (approved):

PR #440, opened 2023-03-09
DynamoDB queries should use ConsistentRead [RHELDST-16604]

dichn:


exodus_lambda/functions/origin_request.py
```diff
@@ -65,6 +65,7 @@ def definitions(self):
             query_result = self.db_client.query(
                 TableName=table,
                 Limit=1,
+                ConsistentRead=True,
```
rohanpm, line 68: It hardly makes a difference, but *this* query doesn't need the ConsistentRead (on config table), only the other one needs it (content table).

For this query it wouldn't make sense if the read had to be consistent because we wrap it with caching anyway.  So we'd be paying the extra cost to do a consistent read, but then getting eventual consistency anyway by usually looking at our potentially stale cache.

rohanpm (changes requested):

dichn:
Removed the ConsistentRead support on config table.

rohanpm (approved):

PR #443, opened 2023-03-22
Add reftest for covering exodus-cdn origin failover [RHELDST-15250]

emanuelelevo:
This adds a new reftest path for an object not being available on the default origin (404 error) while being available on the replica origin.

### DIFF
--- a/support/reftest/data.yml
+++ b/support/reftest/data.yml
@@ -77,3 +77,9 @@ test_data:
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/rhel-8.2-x86_64-boot.iso
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/does-not-exist.iso
   state: absent
+
+# origin content failover
+- path: /_exodus-replica-dummy-object.txt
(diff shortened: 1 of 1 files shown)


negillett (approved):

support/reftest/data.yml
```diff
@@ -77,3 +77,7 @@ test_data:
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/rhel-8.2-x86_64-boot.iso
 - path: /content/dist/rhel8/8.2/x86_64/baseos/iso/does-not-exist.iso
   state: absent
+
+# origin content failover
+- path: /_exodus-replica-dummy-object.txt
```
rohanpm, line 82: Note it's also possible to add the checksum in here which would enable some more tests.  Up to you.

rohanpm (approved):

support/reftest/data.yml
emanuelelevo: Good idea, this file won't change often.

rohanpm (approved):

PR #444, opened 2023-03-22
Improve origin response logging [RHELDST-16814]

negillett:
Adds additional messages for updated request and response, remove json dumps from existing request/response message, leaving formatting to logging handler.

### DIFF
--- a/exodus_lambda/functions/origin_response.py
+++ b/exodus_lambda/functions/origin_response.py
@@ -1,4 +1,3 @@
-import json
 import os
 from base64 import b64encode
 
@@ -18,12 +17,10 @@ class OriginResponse(LambdaBase):
         response = event["Records"][0]["cf"]["response"]
 
         self.logger.debug(
-            "Original request value for origin_response: %s",
-            json.dumps(request, indent=4, sort_keys=True),
+            "Original request value for origin_response: %s", request
         )
         self.logger.debug(
-            "Original response value for origin_response: %s",
-            json.dumps(response, indent=4, sort_keys=True),
+            "Original response value for origin_response: %s", response
         )
 
         if "headers" in request and "want-digest" in request["headers"]:
@@ -50,6 +47,13 @@ class OriginResponse(LambdaBase):
 
         if original_uri:
             self.set_cache_control(original_uri, response)
+
+        self.logger.debug(
+            "Updated request value for origin_response: %s", request
+        )
+        self.logger.debug(
+            "Updated response value for origin_response: %s", response
+        )
 
         return response
 
--- a/exodus_lambda/functions/origin_response.py
+++ b/exodus_lambda/functions/origin_response.py
@@ -17,10 +17,8 @@ class OriginResponse(LambdaBase):
         response = event["Records"][0]["cf"]["response"]
 
         self.logger.debug(
-            "Original request value for origin_response: %s", request
-        )
-        self.logger.debug(
-            "Original response value for origin_response: %s", response
+            "Incoming event for origin_response",
+            extra={"request": request, "response": response},
         )
 
         if "headers" in request and "want-digest" in request["headers"]:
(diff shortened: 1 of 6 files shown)


negillett:
If something in origin response is causing a crash, we'll now be able to tell from at least the lack of "Updated" log messages.

exodus_lambda/functions/origin_response.py
```diff
@@ -18,12 +17,10 @@ def handler(self, event, context):
         response = event["Records"][0]["cf"]["response"]
 
         self.logger.debug(
-            "Original request value for origin_response: %s",
-            json.dumps(request, indent=4, sort_keys=True),
+            "Original request value for origin_response: %s", request
```
rohanpm, line 20: Could you clarify what was your intent with dropping the json.dumps?

I'm just asking in case it was misunderstood that the json.dumps here was redundant after we added json_logging.py.  It's not, because the message part of the log record gets formatted (with its arguments) before it's ever passed into the installed logging.Formatter.  So by removing the json.dumps here, we're now just using the default python string formatting.

This can be seen from the test results where the strings end up formatted with single quotes, which is not valid JSON.

Removing the json.dumps is not necessarily broken but might be less readable since there won't be any indent, possibly unstable since there's no longer any sort_keys, and harder to parse in Splunk.

rohanpm: I see now the JIRA issue RHELDST-16814 says "the current output is invalid JSON / incompatible with splunk", I'm rather confused by that.

If the current output (for each log record) is not valid JSON, that'd be a bug in our JsonFormatter class.  Nothing you can put *inside* the log message should be able to make it produce invalid JSON.

I had a play around in Splunk myself and the current logs *are* valid JSON and are parseable using Splunk's JSON support.  Although, because it's JSON-within-JSON, it does not happen automatically - it has to be explicitly extracted.  

[index="aws-cloudfront" sourcetype="aws:firehose:json" message="Original request value for origin_response:*" | rex field=message "(?s)^[^:]+: (?<req_raw>.+)$" | spath input=req_raw output=s3_uri uri | spath input=req_raw output=original_uri headers.exodus-original-uri{0}.value | top original_uri s3_uri](https://url.corp.redhat.com/a8d344c)

That query uses the current embedded JSON to make a table with input URIs in one column and the S3 URIs in another.  The "spath" expressions are the part which parse the inner JSON.

I would expect dropping the json.dumps to make queries like the above strictly *more* difficult, not less.

rohanpm: I'll also note that there is another way to handle this which would probably make the logs work even better in Splunk: don't embed the objects in the message at all.  Put them in "extra" and let the JsonFormatter dump them, same as it does for other fields like aws-request-id.

```diff
diff --git a/exodus_lambda/functions/json_logging.py b/exodus_lambda/functions/json_logging.py
index c550ef4..f27d4a7 100644
--- a/exodus_lambda/functions/json_logging.py
+++ b/exodus_lambda/functions/json_logging.py
@@ -10,6 +10,8 @@ class JsonFormatter(logging.Formatter):
             "time": "asctime",
             "aws-request-id": "aws_request_id",
             "message": "message",
+            "request": "request",
+            "response": "response",
         }
         self.datefmt = datefmt or "%Y-%m-%d %H:%M:%S"
         self.dictfmt = None
diff --git a/exodus_lambda/functions/origin_response.py b/exodus_lambda/functions/origin_response.py
index 682ba0f..fa5df78 100755
--- a/exodus_lambda/functions/origin_response.py
+++ b/exodus_lambda/functions/origin_response.py
@@ -18,12 +18,8 @@ class OriginResponse(LambdaBase):
         response = event["Records"][0]["cf"]["response"]
 
         self.logger.debug(
-            "Original request value for origin_response: %s",
-            json.dumps(request, indent=4, sort_keys=True),
-        )
-        self.logger.debug(
-            "Original response value for origin_response: %s",
-            json.dumps(response, indent=4, sort_keys=True),
+            "Incoming event for origin_response",
+            extra={"request": request, "response": response},
         )
 
         if "headers" in request and "want-digest" in request["headers"]:
```

That should make everything be parsed by Splunk automatically and the query I shared earlier could be even simpler (no 'rex' or 'spath' needed I guess).

I have not tested that, and I'm not saying you have to do that for this issue if you don't want, but if you wanted to refactor these messages then removing the json.dumps and using default python string formatting seems like going in the wrong direction to me.  Something like the above seems to make more sense.

rohanpm (approved):

crungehottman (approved):

PR #451, opened 2023-04-13
Update Want-Digest to Want-Repr-Digest [RHELDST-17582]

negillett:


tests/integration/test_exodus.py
```diff
@@ -134,19 +134,19 @@ def test_content_type_header_HEAD(
 ]
 
 
-# use Want-Digest/Digest to check if alias take effect
+# use Want-Repr-Digest/Repar-Digest to check if alias take effect
```
rohanpm, line 137:
```suggestion
# use Want-Repr-Digest/Repr-Digest to check if alias take effect
```

rohanpm (dismissed):
Note that pipeline-test will also need an update for this to get through exodus-cdn-deploy.

crungehottman (dismissed):

negillett:
Will squash

negillett:
> Note that pipeline-test will also need an update for this to get through exodus-cdn-deploy.

Thanks for pointing this out!

crungehottman (approved):

PR #454, opened 2023-04-25
Logging: record milliseconds in timestamp by default [RHELDST-16619]

negillett:
Overrides logging's formatTime to use datetime objects and a period
separator for msec. Removes the datefmt set in the lambda config template
but the field remains configurable should the client/service choose.
All log times are now in UTC timezone.

### DIFF
--- a/tests/functions/test_base.py
+++ b/tests/functions/test_base.py
@@ -1,7 +1,10 @@
+import copy
+import json
 import logging
 import os
 
 import pytest
+from freezegun import freeze_time
 
 from exodus_lambda.functions.base import LambdaBase
 
@@ -9,6 +12,7 @@ from ..test_utils.utils import generate_test_config
 
 CONF_FILE = os.environ.get("EXODUS_LAMBDA_CONF_FILE")
 TEST_CONF = generate_test_config()
+MOCKED_DT = "2023-04-26 14:43:13.570034+00:00"
 
 
 def test_base_handler():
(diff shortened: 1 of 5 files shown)


tests/functions/test_base.py
```diff
@@ -69,3 +71,18 @@ def test_json_handler_stack_info(caplog):
...
+    conf = copy.deepcopy(TEST_CONF)
+    conf["logging"]["formatters"]["default"][
+        "datefmt"
+    ] = "%Y-%m-%d %H:%M:%S.%f"
+
+    # Logging shouldn't raise an error
+    LambdaBase(conf_file=conf).logger.info("Works!")
+    # Logged timestamp should be expected length
+    assert len(json.loads(caplog.text)["time"]) == len(
+        "2023-04-25 15:25:51.364875"
```
rohanpm, line 87: I'm having some trouble following this test.  Maybe two related issues.

- what's the "%f" mean in the configured `datefmt` above?  It's not one of the documented `strftime` directives.
- how is this the expected length?  The `.364875` here isn't milliseconds since that should have 3 digits only.

It makes me a little suspicious that the test might pass for the wrong reason.

Is it possible to freeze the time and then check that the formatted timestamp exactly matches an expected value, rather than checking that the length matches the length of some other string?

negillett: This test is just demonstrating that datefmt is configurable. So it edits the configuration to an arbitrary format different than the defaut. The logger's timeFormat now uses datetime, not time, so the `%f` seen here is datetime's microseconds. I understand the confusion, though, as that's pretty close to the new default. I'll update it to make it more distinct and add some more comments.

rohanpm (approved):

dichn (approved):

PR #456, opened 2023-05-01
More selectively resolve aliases

negillett:
Previously, URIs were checked against all origin, rhui and releasever aliases regardless of whether or not they contained the operative segment. Now it will only iterate through origin and rhui aliases if the URI has match potential. No conditional was added for releasever, as conventions are more varied and complex.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -165,11 +165,12 @@ class OriginRequest(LambdaBase):
 
     def resolve_aliases(self, uri):
         # aliases relating to origin, e.g. content/origin <=> origin
-        uri = self.uri_alias(uri, self.definitions.get("origin_alias"))
+        if "origin" in uri:
+            uri = self.uri_alias(uri, self.definitions.get("origin_alias"))
 
         # aliases relating to rhui; listing files are a special exemption
         # because they must be allowed to differ for rhui vs non-rhui.
-        if not uri.endswith("/listing"):
+        if not uri.endswith("/listing") and "rhui" in uri:
             uri = self.uri_alias(uri, self.definitions.get("rhui_alias"))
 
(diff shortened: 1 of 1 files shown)


exodus_lambda/functions/origin_request.py
```diff
@@ -165,11 +165,12 @@ def uri_alias(self, uri, aliases):
 
     def resolve_aliases(self, uri):
         # aliases relating to origin, e.g. content/origin <=> origin
-        uri = self.uri_alias(uri, self.definitions.get("origin_alias"))
+        if "origin" in uri:
```
rohanpm, line 168: I'm having a hard time seeing that there could be any non-negligible improvement from this.

If `definitions` is already loaded, I'd expect the call to uri_alias to be blazing fast.  Probably about the same speed as `if "origin" in uri` , i.e. so close to instant that you can barely even measure it.

If `definitions` is not already loaded, then this does look like it could have value in that it can possibly avoid the DynamoDB query to exodus-config in order to load definitions.  Which could indeed save significant time.

But: that's ruined by the fact that we don't have any similar logic for releasever.

So if I'm reading this right, if resolve_aliases is called, `self.definitions` has to be called 100% of the time - no way around it.  It seems like the only possible speedup here is then avoiding 1-2 calls to `uri_alias` but is that really making any kind of a difference?  I'm very surprised if so.

Do you have some logs somewhere you could share related to this?  I think we should only do this if we have some evidence that avoiding calls to uri_alias would be helpful.

negillett: Yeah, looking at the data I collected, most times were >= 0.02 seconds. This probably won't have any noticeable impact. The average I got was influenced by the occasional query for definitions and, you're right, it's going to query for releasever anyway. Maybe it'd be worth it when those lists are extremely long sometime in the distant future.

PR #461, opened 2023-05-18
reftest: support publishing via exodus-gw

rohanpm:
The reftest script previously worked only by accessing S3 and DynamoDB directly. We really ought to avoid that as much as possible, especially on stage and production environments. Ideally we should not even have the necessary credentials for that.

Let's support deploying reftest data via exodus-gw. That's safer, will be logged appropriately and so on.

The primary motivation here is to make it possible to more safely provision reftest data to newly deployed stage and production environments, though a few other changes relating to safety were made as well:

- don't deploy "absent" items - there is no need to, and it kinda defeats the purpose. e.g. the reftest entry for "does-not-exist.iso" is supposed to test what happens if you query content which doesn't exist... not if you query content which has been explicitly marked absent, which is subtly different.

- don't deploy exodus-config by default. I don't think this should have been included in the default reftest steps. Most of the reftest data represents some invariants which should always be true and should be safe to deploy on a live environment. exodus-config is different in that the config is expected to change over time, and deploying a wrong config to a live environment can break the CDN completely.

- don't accept a --release-date argument. exodus-gw doesn't support it, and I see no reason to let the user control this anyway; there is no use-case for deploying reftest data in the future, and allowing the user to provide this argument opens the possibility of them providing an invalid argument (e.g. timestamps of the incorrect format, or strings which are not valid timestamps at all).

### DIFF
--- a/support/reftest/reftest
+++ b/support/reftest/reftest
@@ -12,16 +12,25 @@ import json
 import os
 import sys
 import tempfile
-from datetime import datetime
+import time
+from datetime import datetime, timezone
+from typing import Optional
 
 import boto3
 import requests
 import yaml
+from botocore.config import Config
+from botocore.exceptions import ClientError
 from tqdm import tqdm
 
 DATA_DIR = os.path.dirname(__file__)
 
 ENDPOINT_URL = os.environ.get("EXODUS_AWS_ENDPOINT_URL") or None
+
+
+def die(message: str):
+    print(message, file=sys.stderr)
+    sys.exit(4)
 
 
 class DBHandler:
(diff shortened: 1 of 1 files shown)


support/reftest/reftest
```diff
@@ -255,15 +520,15 @@ def download_to_local(url, key_path, cert_path, cacert_path):
...
     for item in config.test_data:
         if not item.get("deploy", True):
             # skip items that specify deploy=false
             continue
 
         if item.get("state") == "absent":
-            # skip all the downloading and checking when absent=true
-            db_client.put_absent_item(item["path"])
+            # Just don't put any item.
             continue
```
negillett, line 532: IIRC, we were putting an "absent" item on the db to test that it wasn't retrievable. Will that functionality remain covered?

negillett (approved):

support/reftest/reftest
```diff
@@ -119,6 +128,241 @@ class S3Handler:
...
+    passed by the user.
+    """
+    if opts.bucket or opts.table or opts.config_table:
+        # AWS mode.
+        if opts.gw_url or opts.gw_env:
+            die(
+                (
+                    "You provided both AWS and exodus-gw arguments. "
+                    "Provide only one or the other."
+                )
```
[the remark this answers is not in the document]
rohanpm: Thanks, you're right, fixed it.

support/reftest/reftest
rohanpm: I don't think this particular scenario:

- If an item exists in DB with state 'absent', trying to fetch it gives a 404

...is covered by reftest data now.  I don't think it's a problem though.  It was never my intent for reftest data to cover that.  It's basically just supposed to declare the expected state of the CDN, not declare what should be deployed in test environments.

The only absent item is:

```
- path: /content/dist/rhel8/8.2/x86_64/baseos/iso/does-not-exist.iso
  state: absent
```

...which was added by me, and when adding it I definitely had in mind that I'm trying to declare "this is a URL which we know doesn't contain any content" (therefore fetching it should give 404).  And not "this is a URL which should have an 'absent' item explicitly provisioned in each environment" (therefore fetching it should give 404, but for a different reason).

negillett (approved):

crungehottman (approved):

PR #479, opened 2023-06-27
Log exodus-lambda cold starts [RHELDST-18674]

crungehottman:
(I've left more details in RHELDST-18674)

### DIFF
--- a/exodus_lambda/functions/base.py
+++ b/exodus_lambda/functions/base.py
@@ -44,6 +44,7 @@ class LambdaBase(object):
                 formatter = JsonFormatter(datefmt=datefmt)
                 root_logger.handlers[0].setFormatter(formatter)
             self._logger = logging.getLogger(self._logger_name)
+            self._logger.info("Initializing logger...")
         return self._logger
 
     @property
--- a/tests/functions/test_base.py
+++ b/tests/functions/test_base.py
@@ -80,15 +80,26 @@ def test_json_logger_timestamp(caplog):
     LambdaBase(conf_file=TEST_CONF).logger.info("Works!")
 
     # Logged timestamp should show milliseconds by default
-    assert json.loads(caplog.text) == {
-        "level": "INFO",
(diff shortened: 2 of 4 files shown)


rohanpm (commented):
Some of the tests need updating.

Also, I'm curious when exactly does this message get logged?  It seems like it will be logged as a side-effect of the first time `self.logger` is called.  Is that actually when the lambda is started or is it when the lambda is handling its first event?

Probably not much difference in the typical case, but I'm wondering if there are cases where the lambda might be preemptively started (i.e. python interpreter spawned and code imported) but then a significant amount of time passes before any event is passed into the lambda.

I think the change is fine regardless of the answer to that question.

negillett (approved):

crungehottman:
> Some of the tests need updating.
> 
> Also, I'm curious when exactly does this message get logged? It seems like it will be logged as a side-effect of the first time `self.logger` is called. Is that actually when the lambda is started or is it when the lambda is handling its first event?
> 

It is logged as a side-effect of the first time  `self.logger` is called, which coincides with the lambda handling its first event (not technically when the lambda is started), so it isn't perfect. (But as I mentioned in Jira comments, I don't really see a way around this, as logs from our exodus-lambda logger seem to be the only logs sent to Splunk.)

> Probably not much difference in the typical case, but I'm wondering if there are cases where the lambda might be preemptively started (i.e. python interpreter spawned and code imported) but then a significant amount of time passes before any event is passed into the lambda.

Yeah, I suppose this could be the case. Though, I haven't noticed evidence of preemptive starts (for example, when deploying a new version of exodus-lambda to my dev env - not sure when the lambda might otherwise preemptively start). AWS logs an "INIT Runtime Version: python:3.9.v23 [...]" message each time the lambda is started (this log is visible using CloudWatch); this INIT log isn't produced when a new version of exodus-lamdba is deployed, for example.

> 
> I think the change is fine regardless of the answer to that question.

rohanpm (approved):

PR #505, opened 2023-09-29
Upgrade exodus-cdn to Python3.11 runtime [RHELDST-18509]

crungehottman:


crungehottman:
There seems to be a bug in pip [0], which results in the incorrect handling of extra dependencies. This only seems to be an issue in python 3.11.

```
ERROR: In --require-hashes mode, all requirements must have their
versions pinned with ==. These do not:
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 572.7/572.7 kB 86.5 MB/s eta 0:00:00
    tomli from
https://files.pythonhosted.org/packages/97/75/10a9ebee3fd790d20926a90a2547f0bf78f371b2f13aa822c759680ca7b9/tomli-2.0.1-py3-none-any.whl
(from coverage[toml]==6.5.0->-r test-requirements.txt (line 173))
```

When trying to work around this issue by manually pinning the tomli package (adding tomli to test-requirements.in), an identical `ERROR: In --require-hashes mode, all requirements must have their
versions pinned with ==. These do not:` error crops up for yet another package. 

So, I don't think we can pin exodus-lambda's minimum python version to `>=3.11`, or run pip-compile with python 3.11 unless we want to deal with (maintain messy workarounds for) this pip bug. But, running exodus-lambda in the python 3.11 lambda runtime (with the changes suggested here) seems to work well, despite this:

```
[crungeho@localhost ~]$ curl -H "Cookie: $COOKIE" -H "X-Exodus-Query: 1" "$URL/cookie-test" -v
[...redacted...]
< content-length: 0
< server: CloudFront
< date: Fri, 29 Sep 2023 18:30:28 GMT
< x-exodus-version: 4eb3f4d94b365ea70dc6da36e0e1babb2c6f8b5e
< x-cache: Miss from cloudfront
[...]
```

logs from my origin request lambda (it is no longer crashing with the error mentioned in my exodus-cdn-playbooks MR):

```
INIT_START Runtime Version: python:3.11.v13	Runtime Version ARN: arn:aws:lambda:us-east-1::runtime:[redacted]
[...]
{
    "level": "INFO",
    "logger": "origin-request",
    "time": "2023-09-29 18:30:28.326",
    "aws-request-id": "[redacted]",
    "message": "Querying '[redacted: my dev pre table]' table for '/cookie-test/.__exodus_autoindex'...",
    "request": null,
    "response": null
}
{
    "level": "INFO",
    "logger": "origin-request",
    "time": "2023-09-29 18:30:28.266",
    "aws-request-id": "[redacted]",
    "message": "No item found for URI: /cookie-test",
    "request": null,
    "response": null
}
[...]
```

[0] https://github.com/pypa/pip/issues/9644

rohanpm:
> There seems to be a bug in pip [0], which results in the incorrect handling of extra dependencies. This only seems to be an issue in python 3.11.
> 

How do you reproduce the errors discussed here?

I tried this:

1. Checked out your change 4eb3f4d94b365ea

2. Added this on top:

```
diff --git a/tox.ini b/tox.ini
index 5202dc9..78c79aa 100644
--- a/tox.ini
+++ b/tox.ini
@@ -133,7 +133,7 @@ relative_files = true
 
 [testenv:pip-compile]
 deps = pip-tools
-basepython = python3.9
+basepython = python3.11
 skip_install = true
 skipsdist = true
 commands =
```

3. Ran `tox -e pip-compile`.  It succeeded.

4. Ran some other tox envs: `py311`, `package`.  They succeeded.

I would much prefer if we can solve this, because if we move forward with running pip-compile using python 3.9 but deploying with 3.11, it might initially work but then break without warning at any time.  I see it as tech debt.

I feel like there will be a reasonable solution to this, but the first step would be to reproduce the issue, which I can't so far.

rohanpm (dismissed):
I think we do need to get pip-compile working on python3.11 if at all feasible.

crungehottman:
> I feel like there will be a reasonable solution to this, but the first step would be to reproduce the issue, which I can't so far.

Hmm, that's odd. I pushed the py3.11 pip-compile (and related changes) to show you that the bandit step of the CI fails, but the error has disappeared. I can't see the past CI job logs, so I don't have any evidence to help determine why bandit no longer fails.

Regardless, it is working as expected now:

```
[crungeho@localhost ~]$ curl -H "Cookie: $COOKIE" -H "X-Exodus-Query: 1" "$URL/cookie-test" -v
[...output trimmed...]
< HTTP/2 404 
< content-length: 0
< server: CloudFront
< date: Mon, 02 Oct 2023 16:59:25 GMT
< x-exodus-version: b8df05baa9da1b99d2d963bb76fe143f605cadb0
[...output trimmed...]
```

```
INIT_START Runtime Version: python:3.11.v13	Runtime Version ARN: arn:aws:lambda:us-east-1::runtime:[redacted]
{
    "level": "INFO",
    "logger": "origin-request",
    "time": "2023-10-02 16:59:23.710",
    "aws-request-id": "[redacted]",
    "message": "Initializing logger...",
    "request": null,
    "response": null
}
[...]
```

negillett (approved):

rohanpm (approved):

PR #511, opened 2023-10-13
Reuse policy,signature from cookie request querystring [RHELDST-18071]

negillett:
When a client provides a policy and signature with a cookie request, reuse it when constructing a redirect response to avoid unecessarily loading the cookie key, building a policy, and generating a signature.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -4,6 +4,7 @@ import os
 import time
 import urllib
 from datetime import datetime, timedelta, timezone
+from urllib.parse import parse_qs
 
 import boto3
 import cachetools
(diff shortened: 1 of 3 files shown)


exodus_lambda/functions/origin_request.py
```diff
@@ -172,23 +173,41 @@ def resolve_aliases(self, uri):
...
-            append="; Secure; HttpOnly; SameSite=lax; Domain=%s; Path=/origin/; Max-Age=%s"
-            % (domain, int(ttl.total_seconds())),
-            resource="https://%s/origin/*" % domain,
+        signer = Signer(cookie_key, self.conf.get("key_id"))
+
+        if "content" in uri:
+            path = "content"
+        if "origin" in uri:
+            path = "origin"
+
```
negillett, line 204: I couldn't recall why we were creating cookies for both origin and content paths. Especially when we know the requested URL. So, if that's really necessary, I can revert this bit.

exodus_lambda/functions/origin_request.py
```diff
@@ -172,23 +173,41 @@ def resolve_aliases(self, uri):
     def handle_cookie_request(self, event):
         now = datetime.utcnow()
         ttl = timedelta(minutes=self.conf.get("cookie_ttl", 720))
-        domain = event["Records"][0]["cf"]["config"]["distributionDomainName"]
-        uri = event["Records"][0]["cf"]["request"]["uri"]
+        record = event["Records"][0]["cf"]
+        domain = record["config"]["distributionDomainName"]
+        uri = record["request"]["uri"]
+        querystring_dict = parse_qs(record["request"]["querystring"])
+        policy = querystring_dict.get("Policy", [None])[0]
```
rohanpm, line 180: I don't think these are going to work.  Per https://github.com/release-engineering/exodus-gw/pull/606#discussion_r1359954369 I believe we'll need to have three policies in play here.  If this code here works at all, it's currently picking the policy for `/_/cookie` which is not going to be reusable for any other path, so putting it into a cookie isn't helpful.

I would expect here that we need to pick out the policies for the cookie from a new query string element and not Policy/Signature which was already used to enable access to `/_/cookie`.

exodus_lambda/functions/origin_request.py
```diff
@@ -172,23 +173,41 @@ def resolve_aliases(self, uri):
...
+        signature = querystring_dict.get("Signature", [None])[0]
 
         self.logger.info("Handling cookie request: %s", uri)
 
-        signer = Signer(self.cookie_key, self.conf.get("key_id"))
+        # Only load the cookie key if we need to generate a signature.
+        cookie_key = None
+        if not all([policy, signature]):
```
rohanpm, line 187: I don't think this should be conditional, as one of the major benefits of this change should be that we can completely drop the related secret (and code) from exodus-lambda, but we can't do that as long as we'd still need it in some cases.

If you had in mind that this would be used temporarily, I'd rather we just get the exodus-gw side completed first.  So, let's say exodus-gw does:

- uses a valid signed URL for `/_/cookie` (as it does today)
- but also includes extra policies for /content/ and /origin/ within that request, under a new query string parameter

That should work fine against exodus-cdn today, and the extra policies are just ignored.  Then exodus-cdn can have a single update which makes it start using the extra policies instead of generating them by itself.  We can rip out the signing/secret code in exodus-cdn immediately as part of that change, there doesn't seem to be any need to keep it around as a conditional if we have already updated exodus-gw everywhere.

rohanpm (changes requested):

PR #516, opened 2023-10-19
Cookie Requests: Use forwarded cookies, don't generate [RHELDST-18071]

negillett:
Now that exodus-gw provides the necessary cookie data for redirect requests, origin_request can use those rather than generate them itself.

This commit makes the necessary changes for that and cleans up now-unused code.


crungehottman (approved):

docs/schemas/raw/lambda_config.yaml
```diff
@@ -22,12 +22,6 @@ properties:
       How long, in minutes, to cache config loaded from the config_table
     maximum: 10000
     minimum: 0
-  cookie_ttl:
```
rohanpm, line 25: I'm not sure it's the best idea to immediately remove these from the schema.  That will cause any existing config files to be considered invalid and will fail deployment.  At the very least, it means an update to exodus-cdn-playbooks will have to be deployed *exactly* with this code update - no earlier and no later - because old config won't be accepted by the new schema and vice-versa.

You will probably find that deployment goes more smoothly if you allow these to stay in the schema for now, even if the code no longer uses these config elements.  They can be cleaned up after everything is fully deployed and exodus-cdn-playbooks no longer tries to set these.

I think you'd want to let them stay in "properties" (which means putting them in config will be permitted), but remove them from "required" (which means omitting them in config is also permitted).  That'll give one version of the schema/code which supports both the currently deployed and proposed new config files.

tests/functions/test_origin_request.py
```diff
@@ -609,83 +649,46 @@ def test_origin_request_cookie_uri(
     ]
     assert request["headers"]["set-cookie"] == [
         {
-            "value": "CloudFront-Key-Pair-Id=K1MOU91G3N7WPY; Secure; "
-            "HttpOnly; SameSite=lax; Domain=ex.cloudfront.net; "
-            "Path=/content/; Max-Age=43200"
+            "value": "CloudFront-Key-Pair-Id=XXXXXXXXXXXXXX; Secure; "
+            "HttpOnly; SameSite=lax; Domain=http://localhost:8049; "
+            "Path=/content/*; Max-Age=1800"
```
rohanpm, line 654: This is indicating a mistake, the Path element of the cookie should still be just `/content/` and not `/content/*` as there is nothing about wildcards in the standard for cookies.

I think I pointed this out in the exodus-gw code review already so I guess it's just that you'd copied the query string from an earlier version of that code, but it'd still be good to fix this.

tests/functions/test_origin_request.py
```diff
@@ -609,83 +649,46 @@ def test_origin_request_cookie_uri(
     ]
     assert request["headers"]["set-cookie"] == [
         {
-            "value": "CloudFront-Key-Pair-Id=K1MOU91G3N7WPY; Secure; "
-            "HttpOnly; SameSite=lax; Domain=ex.cloudfront.net; "
-            "Path=/content/; Max-Age=43200"
+            "value": "CloudFront-Key-Pair-Id=XXXXXXXXXXXXXX; Secure; "
+            "HttpOnly; SameSite=lax; Domain=http://localhost:8049; "
```
rohanpm, line 653: The Domain changing to include the scheme might be a mistake.  That would probably be a bug in the exodus-gw change rather than here, but I've just noticed it here.

I don't actually know what the spec says, but e.g. at https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies the only example of Domain is `Domain=mozilla.org` - no scheme.

And the behavior we've implemented until now also doesn't include a scheme:

```
$ curl -Lv -b '' https://exodus-gw.dev.psi.redhat.com/rmcgover-pre/cdn/content/waf-test 2>&1 | grep -i Set-Cookie | tail -n1
< set-cookie: CloudFront-Signature=[...]; Secure; HttpOnly; SameSite=lax; Domain=dquugasaklmli.cloudfront.net; Path=/origin/; Max-Age=43200
```

Can you please check this?  We should probably not change this behavior if we don't have a specific reason to, e.g. if you found that a spec says we're doing it wrong.

rohanpm (changes requested):

tests/functions/test_origin_request.py
negillett: Hmm Okay. It is present in exodus-gw as well but can change it there before it's deployed.

tests/functions/test_origin_request.py
negillett: Ah, yeah, sounds good. I'll make that change in gw.

negillett:
Resolved some comments via https://github.com/release-engineering/exodus-gw/pull/612

docs/schemas/raw/lambda_config.yaml
negillett: Restored docs and configuration file template, removed cookie-related settings from required list
Once gw is deployed and config updated a follow-up can be created to remove these

negillett:
Once this is deployed I will update the config deployed by exodus-cdn-playbooks

negillett:
https://github.com/release-engineering/exodus-gw/pull/612 should be merged and deployed before this is submitted.

rohanpm (approved):

tests/functions/test_origin_request.py
```diff
@@ -609,83 +608,46 @@ def test_origin_request_cookie_uri(
...
         {
             "value": "CloudFront-Policy=eyJTdGF0ZW1lbnQiOlt7IlJlc291cmNlIjoiaH"
-            "R0cHM6Ly9leC5jbG91ZGZyb250Lm5ldC9jb250ZW50LyoiLCJDb25kaXRpb24iOns"
-            "iRGF0ZUxlc3NUaGFuIjp7IkFXUzpFcG9jaFRpbWUiOjF9fX1dfQ__; Secure; "
-            "HttpOnly; SameSite=lax; Domain=ex.cloudfront.net; "
-            "Path=/content/; Max-Age=43200"
+            "R0cDovL2xvY2FsaG9zdDo4MDQ5L2NvbnRlbnQvKiIsIkNvbmRpdGlvbiI6eyJEYXR"
+            "lTGVzc1RoYW4iOnsiQVdTOkVwb2NoVGltZSI6MTY0NDk3MTQwMH19fV19; "
+            "Secure; HttpOnly; SameSite=lax; Domain=localhost:8049; "
+            "Path=/content/; Max-Age=1800"
```
rohanpm, line 620: I think this significant reduction of Max-Age is pointing at another issue.  It is not desired that this change should make the cookies expire faster than they used to.

This *could* just be a test data issue, but I think it's a real issue which again has to be fixed on the exodus-gw side.  See: https://github.com/release-engineering/exodus-gw/pull/612#pullrequestreview-1693573305

rohanpm (changes requested):

negillett:
This will need to be rebased onto https://github.com/release-engineering/exodus-lambda/pull/519

negillett:
Rebased

exodus_lambda/functions/origin_request.py
```diff
@@ -171,28 +125,13 @@ def resolve_aliases(self, uri):
...
 
     def handle_cookie_request(self, event):
-        now = datetime.utcnow()
-        ttl = timedelta(minutes=self.conf.get("cookie_ttl", 720))
-        domain = event["Records"][0]["cf"]["config"]["distributionDomainName"]
-        uri = event["Records"][0]["cf"]["request"]["uri"]
+        request = event["Records"][0]["cf"]["request"]
+        uri = request["uri"]
+        params = {k: v[0] for k, v in parse_qs(request["querystring"]).items()}
+        set_cookies = json.loads(cf_b64decode(params["CloudFront-Cookies"]))
```
rohanpm, line 131: Can we please write this in a way so that it doesn't crash when there's a problem with this parameter?  (e.g. it's missing, not valid base64 or not valid JSON.)  We don't want to get lambda execution errors on bad inputs from clients because it'll make it hard to pick out or set up alerts for genuine errors from exodus-lambda.

It should probably also be validated that the JSON loading result strictly gives us a list of string and not anything else.

If any error does happen in this step, I think we should return a 400 (client error) response.

rohanpm (changes requested):

rohanpm (approved):

PR #521, opened 2023-11-02
Reduce boto connect/read timeouts [RHELDST-21242]

negillett:
The default value of connect_timeout and read_timeout in boto is 60 seconds. Our lambdas timeout after 15 seconds. There's no reason for the inner timeouts to exceed the outer and reducing the inner timeouts may prove more helpful in debugging.


negillett:
Hold on this until CDN config changes are in place.

docs/schemas/raw/lambda_config.yaml
```diff
@@ -93,6 +110,8 @@ required:
 - table
 - config_table
 - config_cache_ttl
+- connect_timeout
```
rohanpm, line 113: Note that adding these as "required" means it won't be backwards-compatible with existing config files so it will force an update in the playbooks at the same time.  You might want to consider not adding them to required, or only adding them to required later.

rohanpm (approved):

docs/schemas/raw/lambda_config.yaml
negillett: Yeah, that's why I've made this dependant on the cdn-playbooks change. Once that's fully deployed this will be safe.

PR #614, opened 2024-12-09
Support "exclude_paths" in exodus-cdn [RHELDST-26014]

amcmahon-rh:
Certain paths on our NetStorage fs have no symlinks/aliases, and teams/tools that make use of these paths expect similar behaviour from exodus.

This change introduces "exclude_paths" to exclude alias generation for certain paths, so we can replicate NetStorage behaviour. We maintain the previous behaviour as a "fallback" in case the item key hasn't or can't be updated in the database. This will prevent items from disappearing once we cut over.

### DIFF
--- a/exodus_lambda/functions/origin_request.py
+++ b/exodus_lambda/functions/origin_request.py
@@ -103,11 +103,15 @@ class OriginRequest(LambdaBase):
             processed = []
 
             for alias in remaining:
-                exclusion_match = not ignore_exclusions and \
-                                  any([re.search(exclusion, uri) for exclusion
-                                       in alias.get("exclude_paths",[])])
-                if (uri.startswith(alias["src"] + "/") or uri == alias["src"]) \
-                        and not exclusion_match:
+                exclusion_match = not ignore_exclusions and any(
+                    [
+                        re.search(exclusion, uri)
+                        for exclusion in alias.get("exclude_paths", [])
+                    ]
+                )
+                if (
+                    uri.startswith(alias["src"] + "/") or uri == alias["src"]
+                ) and not exclusion_match:
                     uri = uri.replace(alias["src"], alias["dest"], 1)
                     processed.append(alias)
 
@@ -125,17 +129,20 @@ class OriginRequest(LambdaBase):
     def resolve_aliases(self, uri, ignore_exclusions=False):
         # aliases relating to origin, e.g. content/origin <=> origin
 
-        uri = self.uri_alias(uri, self.definitions.get("origin_alias"),
-                             ignore_exclusions)
+        uri = self.uri_alias(
+            uri, self.definitions.get("origin_alias"), ignore_exclusions
+        )
         # aliases relating to rhui; listing files are a special exemption
         # because they must be allowed to differ for rhui vs non-rhui.
         if not uri.endswith("/listing"):
-            uri = self.uri_alias(uri, self.definitions.get("rhui_alias"),
-                                 ignore_exclusions)
+            uri = self.uri_alias(
+                uri, self.definitions.get("rhui_alias"), ignore_exclusions
+            )
 
         # aliases relating to releasever; e.g. /content/dist/rhel8/8 <=> /content/dist/rhel8/8.5
-        uri = self.uri_alias(uri, self.definitions.get("releasever_alias"),
-                             ignore_exclusions)
+        uri = self.uri_alias(
+            uri, self.definitions.get("releasever_alias"), ignore_exclusions
+        )
 
         self.logger.debug("Resolved request URI: %s", uri)
 
@@ -377,8 +384,9 @@ class OriginRequest(LambdaBase):
             return self.handle_cookie_request(event)
 
         preferred_uri = self.resolve_aliases(request["uri"])
-        fallback_uri = self.resolve_aliases(request["uri"],
-                                            ignore_exclusions=True)
+        fallback_uri = self.resolve_aliases(
+            request["uri"], ignore_exclusions=True
+        )
         uris = [preferred_uri]
         # Some file keys might take a while to update to reflect URI alias exclusions.
         # Allowing the original behaviour as a fallback will avoid a flood of 404 errors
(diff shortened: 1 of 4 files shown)


tests/functions/test_alias.py
```diff
@@ -16,8 +18,8 @@ def test_alias_single():
     req = OriginRequest(conf_file=TEST_CONF)
 
     aliases = [
-        {"src": "/foo/bar", "dest": ""},
-        {"src": "/baz", "dest": "/quux"},
+        {"src": "/foo/bar", "dest": "", "exclude_paths": []},
+        {"src": "/baz", "dest": "/quux", "exclude_paths": []},
```
crungehottman, line 22: Please be sure to test alias resolution for aliases without an `exclude_paths` key. None of the aliases in the current config (deployed to prod) contain an `exclude_paths` key.

tests/functions/test_alias.py
```diff
@@ -43,6 +45,114 @@ def test_alias_equal():
 
     req = OriginRequest(conf_file=TEST_CONF)
 
-    aliases = [{"src": "/foo/bar", "dest": "/quux"}]
+    aliases = [{"src": "/foo/bar", "dest": "/quux", "exclude_paths": []}]
```
crungehottman, line 48: Please be sure to test alias resolution for aliases without an `exclude_paths` key.

tests/functions/test_alias.py
```diff
@@ -32,7 +34,7 @@ def test_alias_boundary():
 
     req = OriginRequest(conf_file=TEST_CONF)
 
-    aliases = [{"src": "/foo/bar", "dest": "/"}]
+    aliases = [{"src": "/foo/bar", "dest": "/", "exclude_paths": []}]
```
crungehottman, line 37: Please be sure to test alias resolution for aliases without an `exclude_paths` key.

tests/test_data/exodus-config.json
```diff
@@ -2,159 +2,197 @@
     "origin_alias": [
         {
             "src": "/content/origin",
-            "dest": "/origin"
+            "dest": "/origin",
+            "exclude_paths": []
```
crungehottman, line 6: I suppose it's OK to leave a couple of aliases with `"exclude_paths": []`, but to accurately reflect the current state of the CDN, please omit the exclude_paths key/pair altogether instead of specifying an empty list (`"exclude_paths": []`), so that we can test that the CDN can handle alias resolutions when an alias does not contain an `exclude_paths` key (i.e., the current state of **all** aliases on the CDN).

tests/test_data/exodus-config.json
```diff
@@ -2,159 +2,197 @@
...
         },
         {
             "src": "/content/rc/rhel/rhui",
-            "dest": "/content/rc/rhel"
+            "dest": "/content/rc/rhel",
+            "exclude_paths": []
         }
     ],
     "releasever_alias": [
```
crungehottman, line 136: Perhaps leave a few `releasever_alias`es without an `exclude_paths` key so we can test that the CDN can handle alias resolutions when an alias does not contain an exclude_paths key (i.e., the current state of all aliases on the CDN).

exodus_lambda/functions/origin_request.py
```diff
@@ -304,33 +319,9 @@ def validate_request(self, request):
...
-
-        listing_response = self.handle_listing_request(uri)
-        if listing_response:
-            self.set_cache_control(uri, listing_response)
-            return listing_response
-
-        table = self.conf["table"]["name"]
+    def handle_file_request(self, request, table, original_uri, uri):
+        # Try find the db entry corresponding to the uri. We try the uri, it's
+        # corresponding index path and then the fallback_uri.
```
crungehottman, line 324: I think stating "and then the fallback_uri" is confusing here, because there's no concept of `fallback_uri` within this function.

exodus_lambda/functions/origin_request.py
```diff
@@ -304,33 +319,9 @@ def validate_request(self, request):
...
-        uri = self.resolve_aliases(request["uri"])
-
-        listing_response = self.handle_listing_request(uri)
-        if listing_response:
-            self.set_cache_control(uri, listing_response)
-            return listing_response
-
-        table = self.conf["table"]["name"]
+    def handle_file_request(self, request, table, original_uri, uri):
+        # Try find the db entry corresponding to the uri. We try the uri, it's
```
crungehottman, line 323: s/it's/its

crungehottman (changes requested):

exodus_lambda/functions/origin_request.py
```diff
@@ -374,6 +364,60 @@ def handler(self, event, context):
...
+        original_uri = request["uri"]
+
+        if request["uri"].startswith("/_/cookie/"):
+            return self.handle_cookie_request(event)
+
+        uri = self.resolve_aliases(request["uri"])
+        fallback_uri = self.resolve_aliases(
+            request["uri"], ignore_exclusions=True
+        )
+        fallback_uri = fallback_uri if uri != fallback_uri else None
```
crungehottman, line 390: It may be helpful to include a comment explaining the purpose/context of "fallback_uri" for posterity.

exodus_lambda/functions/origin_request.py
```diff
@@ -374,6 +364,60 @@ def handler(self, event, context):
...
+        if listing_response := self.handle_listing_request(uri):
```
rohanpm, line 391: This is subtly different than I'd expected.

The implemented logic is:

1. try to give a listing response using preferred uri.
2. try to give a listing response using fallback uri.
3. try to give a file response using preferred uri.
4. try to give a file response using fallback uri.
5. give 404 response

The acceptance criteria does not spell it out in detail, but I'd rather expected to see:

1. try to give a listing response using preferred uri.
2. try to give a file response using preferred uri.
3. try to give a listing response using fallback uri.
4. try to give a file response using fallback uri.
5. give 404 response

I thought that would be the natural way to implement the code, as one could take the existing logic and wrap it in a `for uri in (preferred, fallback)` loop.

The way it's been implemented now means that the fallback uri could theoretically be used even if a response is available using the preferred uri. In other words, `exclude_paths` might be ignored even if it were possible for it to be respected.

I'm reluctant to insist on changing this as I think there's a good chance that, given realistic content and config, there will never be a different result between the two approaches anyway.  It seemed worth pointing out this quirk though.

rohanpm (dismissed):

negillett (approved):

rohanpm (approved):

crungehottman (approved):

PR #628, opened 2025-02-05
Introduce mirrored reads [RHELDST-28332]

crungehottman:
This commit modifies exodus-cdn's lookup behavior to enable mirrored reads. This means, that when exodus-cdn looks up content to be served for a path with a $releasever alias in effect, exodus-cdn attempts to look up content on both sides of the alias. (Previously, exodus-cdn only performed a content lookup on the destination side.)

The mirrored reads are enabled by default, but they can be disabled by setting the new config option "disable_mirrored_reads" to true (1, '1', or 'True', where 'True' is not case sensitive.)


configuration/lambda_config.template
```diff
@@ -17,6 +17,7 @@
   "headers": {
     "max_age": $EXODUS_HEADERS_MAX_AGE
   },
+  "disable_mirrored_reads": "$EXODUS_LAMBDA_DISABLE_MIRRORED_READS",
```
rohanpm, line 20: Are you sure you want to have the option with a negative name like this ("disable")?

I know that the option defaults to False (i.e. the feature of mirrored reads defaults to enabled).  However, in the long run I think code usually ends up easier to understand if all feature flags have a consistent positive name (i.e. *enabling* the option *enables* a feature) and you implement whatever default values are desired.

Think of it this way, if you have a set of feature flags then you have a choice between:

1. Set up all the feature flags with positive names, so that *setting* them turns *on* a feature.  Then give some of the flags default values of 0 and others default values of 1, depending on what features you want enabled by default.  Result: consistent names, inconsistent default values.

2. Set up all the feature flags to have the same default value (of 0/unset?).  Then give each flag either a positive or negative name ("enable_foo" or "disable_foo") depending on whether you want the feature to default to on or off. Result: inconsistent names, consistent default values.

This flag now does (2) but I think (1) is much easier to understand.  With (2) the code will end up with a mixture of feature flag handling, some "if enabled_foo" and some "if not disabled_foo".  Also (2) risks becoming inconsistent on both sides over time, because if you change your mind on whether a feature should default to on or off, you're probably not going to go back and rename the feature flag swapping out the negative name with a positive one or vice-versa.  So, in the long run you'll probably end up with neither consistent names, nor consistent default values.

rohanpm (dismissed):

rohanpm (dismissed):

PR #654, opened 2025-06-13
ci: pin pip version in workflow [RHELDST-31877]

dichn:
Pin pip to <=25.0 in GitHub Actions to avoid issues introduced in pip v25.1

### DIFF
--- a/.github/workflows/pip-compile.yml
+++ b/.github/workflows/pip-compile.yml
@@ -15,7 +15,7 @@ jobs:
          python-version: "3.11"
 
      - name: Install tox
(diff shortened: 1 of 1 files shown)


negillett:
@dichn, can you try to shorten the commit subject line? It's cut off and bleeding into the commit message.
Same with https://github.com/release-engineering/exodus-gw/pull/814

negillett:
Hmm. Actually, I think it'd be better to use <=25.1 for all CI rather than just reinstalling it for pip-compile jobs.
But this should work as-is and will be temporary anyway.

